PR #5, opened 2017-06-02
Convert Object variables to maps / arrays.

DrRataplan:
This allows passing most javascript POJO things to the XPath engine,
greatly improving iteropts.


src/evaluateXPath.js
```diff
@@ -19,7 +18,7 @@ import isInstanceOfType from './selectors/dataTypes/isInstanceOfType';
  *  * Else, the sequence is atomized and returned.
  *
  * @param  {!string}       xpathSelector  The selector to execute. Supports XPath 3.1.
- * @param  {!Node}         contextNode    The node from which to run the XPath.
+ * @param  {*|null}        contextNode    The node from which to run the XPath.
```
bwrrp: rename to contextItem?

src/selectors/adaptJavaScriptValueToXPathValue.js
```diff
@@ -1,39 +1,50 @@
 import Sequence from './dataTypes/Sequence';
 import createAtomicValue from './dataTypes/createAtomicValue';
 import NodeValue from './dataTypes/NodeValue';
+import ArrayValue from './dataTypes/ArrayValue';
+import MapValue from './dataTypes/MapValue';
 
 function adaptItemToXPathValue (value) {
```
bwrrp: Now returns a Sequence, rename to reflect this?

bwrrp: Also, none of these branches actually returns more than a singleton sequence. Why not leave the Sequence constructor call where it was?

src/selectors/path/PathSelector.js
```diff
@@ -162,22 +162,34 @@ class PathSelector extends Selector {
...
-			let resultValuesInOrderOfEvaluation = /** @type {!IteratorIterable<!../dataTypes/Sequence>} */ ({
+			let resultValuesInOrderOfEvaluation = /** @type {!IteratorIterable<!Sequence>} */ ({
 				[Symbol.iterator]: () => resultValuesInOrderOfEvaluation,
 				next: () => {
 					const childContext = childContextIterator.next();
 					if (childContext.done) {
 						return { done: true };
 					}
-					if (!isInstanceOfType(childContext.value.contextItem, 'node()')) {
+					if (childContext.value.contextItem !== null && !isInstanceOfType(childContext.value.contextItem, 'node()')) {
```
bwrrp: You could consider having isInstanceOfType always return false for null

src/selectors/path/PathSelector.js
```diff
@@ -162,22 +162,34 @@ class PathSelector extends Selector {
...
+				let hasPassedFirst = false;
+				childContextIterator = {
+					next: () => {
+						if (hasPassedFirst) {
+							return { done: true };
+						}
+						hasPassedFirst = true;
+						return { done: false, value: dynamicContext };
+					}
+				};
```
bwrrp: Extract createSingleValueIterator function?

src/selectors/path/PathSelector.js
DrRataplan: I'd say that should be an error, absent context is different from wrong context.

PR #23, opened 2018-02-28
Map external variables with the value null to the empty sequence

### LINKED ISSUE #22: An error is thrown when an object containing null values is inputted as an external variable
```
// This throws an error
evaluateXPathtoBoolean('$x("x") => empty', null, undefined, {x: {x: null}});
```

Let's make 'null' analogous to the empty sequence, like we do if we explicitly convert a JavaScript null to 'item()?'.

DrRataplan:
This fixes #22


src/selectors/adaptJavaScriptValueToXPathValue.js
```diff
@@ -5,6 +5,12 @@ import MapValue from './dataTypes/MapValue';
 import createNodeValue from './dataTypes/createNodeValue';
 import { trueBoolean, falseBoolean } from './dataTypes/createAtomicValue';
 
+/**
+ * Adapt a JavaScript value to the equivalent in XPath. This dynamically assigns the closest type
+ *
+ * @param  {?} value
+ * @return {?./dataTypes/Value} Null if the value is absent and a singleton sequence should be outputted instead
```
bwrrp: s/singleton/empty/ ?

src/selectors/adaptJavaScriptValueToXPathValue.js
```diff
@@ -14,23 +20,44 @@ function adaptItemToXPathValue (value) {
...
 			}
 			if (Array.isArray(value)) {
-				return new ArrayValue(value.map(val => Sequence.singleton(adaptItemToXPathValue(val))));
+				return new ArrayValue(value.map(
+					arrayItem => {
+						const adaptedArrayItem = adaptItemToXPathValue(arrayItem);
+						if (adaptedArrayItem === null) {
+							return Sequence.empty();
+						}
+						return Sequence.singleton(adaptedArrayItem);
```
bwrrp: you could extract an `adaptItemToXPathSingletonOrEmptySequence` function here

bwrrp: Tests seem to be failing with:

```
Chromium 58.0.3029 (Ubuntu 0.0.0) ERROR
  {
    "message": "Some of your tests did a full page reload!",
    "str": "Some of your tests did a full page reload!"
  }
```

PR #24, opened 2018-03-02
Optimize intersect/except to not be quadratic

DrRataplan:
It turns out except is used extensively over large sets, so the naive implementation is
insufficient.
Let's optimize it by not implementing it as a function, but as a variant on merge-sort.


src/domBackedDomFacade.js
```diff
@@ -2,33 +2,39 @@
...
  * @implements {IDomFacade}
  */
-function ReadOnlyDomFacade () {
+function DomBackedDomFacade () {
+	/**
+	 * Defines the ordering of detached nodes, to ensure stable sorting of unrelated nodes.
+	 *
+	 * @type {!Array<!./selectors/dataTypes/Value>}
+	 */
+	this.orderOfDetachedNodes = [];
```
bwrrp: This may leak memory, unless we set its length to 0 after each evaluateXPath, or wrap DomBackedDomFacade in a temporary DomFacade.

src/selectors/operators/IntersectExcept.js
```diff
@@ -0,0 +1,147 @@
...
+ * @param  {!IDomFacade}  domFacade
+ * @param  {!Sequence}  sequence
+ * @param  {number}     expectedResultOrder
+ *
+ * @return {!Sequence}
+ */
+function ensureSortedSequence (intersectOrExcept, domFacade, sequence, expectedResultOrder) {
+	sequence = sequence.map(value => {
+		if (!isSubtypeOf(value.type, 'node()')) {
+			throw new Error(`XPTY0004: Sequences inputted to ${intersectOrExcept} should only contain nodes.`);
```
bwrrp: s/inputted/given/

src/domBackedDomFacade.js
DrRataplan: The domFacade is now always wrapped.

bwrrp: Does that mean that this field is now unused (and can be removed)?

PR #26, opened 2018-03-15
Fix variables object mutation

ThomasBrekelmans:
The variables object is given a fake/easter-egg/test variable like this: { 'theBest': 'FontoXML is the best!' }. But that is Object assigned to the input variables object, which mutates it. This causes issues in FxXPath's componentWillReceiveProps check.

You forgot to Object.assign this into a new object. This edit fixes this.


DrRataplan:
Thanks, I've added a test an I've cleaned up the actual implementation since the variables object has a default anyway.

PR #39, opened 2018-06-25
Allow custom functions to return xml date data types.

### LINKED ISSUE #38: Unable to return xs:date or xs:datetime from a custom xpath function
If you register a custom function that returns an xs:date or xs:datetime, you receive the following error at runtime:

> `Values of the type "${type}" can not be adapted to equivalent XPath values.`

Would it be possible to update the case statement to allow dates and datetimes to be returned from custom functions?
Thanks.

ymor:
I wasn't able to complete a run of the qt3 tests as it constantly kept freezing both chrome and firefox at around 31%.  Though as I've not touched any of the functions, I think this should be OK.

I had to make a few changes to get it to build/run the tests on windows.  These changes should also work on linux/mac.

Fixes #38


devatwork:
@ymor, thanks for the contribution! Don't worry about the QT3 tests, our CI will also run them. @ThomasGravekamp could you do a first review since @DrRataplan is on his holiday?

test/qt3tests.js
```diff
@@ -30,12 +30,12 @@ if (!(typeof window === 'undefined')) {
 	}
 	else {
 		shouldRunTestByName = require('text-loader!./runnableTestSets.csv')
-			.split('\n')
+			.split('/\r?\n/')
```
devatwork: @ymor Change looks robust to me. Did you consider using https://nodejs.org/api/os.html#os_os_eol instead?

ymor: Can't say I've really used node before! But that does look like the better solution.

devatwork: I'm no expert either. @bwrrp, @ThomasGravekamp, @DrRataplan recommendations?

bwrrp: Given the mess that end-of-line handling can be (depending on git settings, we could get \n on Windows), I guess it's probably best to keep the regex and accept both types here.

ymor: I have noticed that I didn't add the regex to the "on node" side of the if/else.  Want me to update that too?

bwrrp: Good catch! I guess those would also need to be changed for the tests to run successfully on Windows.

ymor: I'd assume so but I'm not sure how to run the tests purely in node.  Btw that change to the regex you did, means i can actually fully run the qt3 tests locally on windows now. 👍

bwrrp: Good to hear. I think the node-specific code is not currently active (we'd eventually like to run all tests in node as well as the browser). I've updated the splits anyway, to hopefully prevent the issue returning when we enable this.

DrRataplan:
We might have to add support for date-like types as parameters to custom functions, arguments to entire XPaths and as possible return types, but this PR has a lot of value without it!

@ymor, thanks a lot for making this PR! I'll integrate it and publish it somewhere this week!

PR #55, opened 2018-10-02
Bugfix/#54 incorrect calculation value

### LINKED ISSUE #54: Incorrect value from calculation
Hi,
The below example is returning the incorrect value. 
https://xpath.playground.fontoxml.com/?xml=&xpath=5+-+1+-+1

**Expected value**
3

**Actual value**
5

It seems to be doing 5 - (1 - 1)  instead?

ymor:
Fixes #54 
I had a quick search through the unrunnable QT3 test cases but I couldn't find any that would have surfaced this issue.  I've added two simple test cases, happy to add more if required.


DrRataplan:
Wow, cool man!

I'll be sure to look into the PR today! In the mean time, I've filed an issue with the QT3 test set to add a test for this behaviour: https://www.w3.org/Bugs/Public/show_bug.cgi?id=30292. I wonder when they'll be able to act on that.

Thanks!

src/parsing/xpath.pegjs
```diff
@@ -121,8 +121,12 @@ RangeExpr
 
 // 21
 AdditiveExpr
- = lhs:MultiplicativeExpr _ op:("-" / "+") _ rhs:AdditiveExpr {return ["binaryOperator", op, lhs, rhs]}
- / MultiplicativeExpr
+ = lhs:MultiplicativeExpr rest:( _ op:("-" / "+") _ rhs:MultiplicativeExpr {return {op: op, rhs: rhs}})* {
+        return rest.length === 0 ? lhs : rest.reduce(function (inner, nesting) {
```
DrRataplan: The `length === 0` is not actually needed; Array.prototype.reduce returns the initial value if the array is empty. But let's be explicit here

ymor:
Cheers for merging and releasing this so fast! 
Out of interest, when/how does your playground site get updated?

DrRataplan:
I usually update the playground every update, but I forgot it this time. We should really automate it, but having it as a private repo doesn't make automation any easier...

I'll update it now, it'll take a couple of minutes though

PR #57, opened 2018-10-23
XQueryX parser

DrRataplan:


DrRataplan:
It seems like the following QT3 tests are new-failing:

```
Alnilam /tmp/diff » diff -u before after | grep -E "^\+" 
+++ after	2018-10-25 09:55:31.355792683 +0200
+  "comp-attr-bad-name-1",
+  "comp-attr-bad-name-2",
+  "comp-attr-bad-name-3",
+  "comp-attr-bad-name-4",
+  "comp-attr-bad-name-5",
+  "comp-attr-bad-name-6",
+  "comp-attr-bad-name-7",
+  "K2-ComputeConComment-4",
+  "comp-elem-bad-name-2",
+  "comp-elem-bad-name-3",
+  "comp-elem-bad-name-4",
+  "comp-elem-bad-name-5",
+  "comp-elem-bad-name-6",
+  "K2-ConText-5",
+  "K2-DirectConElem-53",
+  "K-FunctionCallExpr-25",
```

DrRataplan:
The problem with most of these failing tests are dependency related. We are not an XQ10/XQ20/XQ30 engine. We are XQ3.1(+)!

PR #62, opened 2018-11-19
Fix Demo Link in Readme.MD

ymor:
The old link - which i assume just did a redirect - is no longer working.


DrRataplan:
Hi Mark,

Thanks! I was meaning to get around to that!

PR #71, opened 2019-01-21
Dev 2287: Map value can only be read once when injected via variables

JosVerburg:
See unit test "can read a map value twice from within a function" from VarRef.tests.ts for the failing case


DrRataplan:
Great!

PR #73, opened 2019-01-21
Implement (optional) stack traces for use in XPath/XQuery

DrRataplan:
By setting the `debugMode` (open to other names), you can get a rudimentary stacktrace from the XPath engine.

They are formatted like this:

For the XQuery:

```xquery
declare function local:prrt ($a) {
  <ele>{zero-or-one(($a, 3))}</ele>
};

local:prrt(1)
``` 

output:

```
Error: FORG0003: The argument passed to fn:zero-or-one contained more than one item.
  at <functionCallExpr>:2:9 - 2:29
  at <elementConstructor>:2:3 - 2:36
  at <functionCallExpr>:6:1 - 6:14
```

This feels like a valuable MVP for tracing. We can look into displaying more information at a later point.

I made this debugging information opt-in; I expect performance issues if we always enable it: there may be a lot of try-catches generated at some point.

TODO: tests.

@wvbe, what would you say? Will this help you while debugging complex XQuery scripts?


wvbe:
@DrRataplan That looks fantastic! Are there any implications for running the new fontoxpath with `debugMode: false`?

DrRataplan:
@wvbe Nope, there are no implications when using the engine 'normally'. The stack trace points are injected just-in-time by the parser, and only in debug mode.

src/evaluateXPath.ts
```diff
@@ -124,13 +128,64 @@ function transformXPathItemToJavascriptObject(value, dynamicContext) {
...
+	const rawLines = selector.replace('\r', '').split('\n');
+	const lines = rawLines.reduce((markedLines: string[], line, i) => {
+		const lineNumber = i + 1;
+
+		// Only mark 2 lines before and after
+		if (errorLocation.start.line - lineNumber > 2 || lineNumber - errorLocation.end.line > 2) {
+			return markedLines;
+		}
+
+		const prefix = `${Array(rawLines.length.toString().length)
```
DrRataplan: We could also get the width of number (prefix) using logarithms: `const linePrefixWidth = Math.floor(Math.log10(Math.min(errorLocation.end.line + 2, rawLines.length))) + 1`. This may be a bit more readable.

PR #76, opened 2019-01-31
Remove all outward usage of ExternalDomFacade

DrRataplan:
IDomFacade is the outward facing interface, which works with the Node type and its subclasses. ExternalDomFacade is the default wrapper which only translates method calls to property gets. IWrappingDomFacade is the internal domFacade, which uses the ConcreteNode type unions and has additional methods/properties.

### DIFF
--- a/src/domFacade/IDomFacade.ts
+++ b/src/domFacade/IDomFacade.ts
@@ -1,20 +1,11 @@
-import {
-	ConcreteAttributeNode,
-	ConcreteCharacterDataNode,
-	ConcreteChildNode,
-	ConcreteElementNode,
-	ConcreteNode,
-	ConcreteParentNode
-} from './ConcreteNode';
-
 export default interface IDomFacade {
-	getAllAttributes(node: ConcreteElementNode): ConcreteAttributeNode[];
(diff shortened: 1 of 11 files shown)


src/domFacade/IDomFacade.ts
```diff
@@ -1,20 +1,11 @@
...
-	getFirstChild(node: ConcreteParentNode): ConcreteChildNode;
-	getLastChild(node: ConcreteParentNode): ConcreteChildNode;
-	getNextSibling(node: ConcreteChildNode): ConcreteChildNode;
-	getParentNode(node: ConcreteNode): ConcreteParentNode;
-	getPreviousSibling(node: ConcreteChildNode): ConcreteChildNode;
+	getAllAttributes(node: Element): Attr[];
+	getAttribute(node: Element, attributeName: string): string | null;
+	getChildNodes(node: Node): Node[];
+	getData(node: Attr | CharacterData): string;
+	getFirstChild(node: Node): Node | null;
```
JosVerburg: What is the reason for now also allowing null to be returned?

DrRataplan: The first child of an element may be null. This was a mistake in the old typings

PR #80, opened 2019-02-11
Implement has-children

mehmetscoskun:


test/specs/parsing/functions/functions.node.tests.ts
```diff
@@ -332,4 +332,26 @@ return $node/root() = $element`,
 			);
 		});
 	});
+
+	describe('has-children()', () => {
+		it.only('yes, it has children', () => {
```
DrRataplan: Could you remove the 'only' clauses for the tests?

DrRataplan: And also, I'm sure you can think of better test names

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -19,7 +21,7 @@ import FunctionDefinitionType from './FunctionDefinitionType';
 
 const fnString = builtinStringFunctions.functions.string;
 
-function contextItemAsFirstArgument(fn, dynamicContext, executionParameters, _staticContext) {
+function contextItemAsFirstArgument(fn: FunctionDefinitionType, dynamicContext, executionParameters, _staticContext) {
```
DrRataplan: Could you run Prettier over this file?

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -86,6 +93,22 @@ const fnName: FunctionDefinitionType = function(
...
+const fnHasChildren: FunctionDefinitionType = function(
+	_dynamicContext,
+	_executionParameters,
+	_staticContext,
+	nodeSequence: ISequence
+) {
+	return zipSingleton([nodeSequence], ([nodeValue]: (Value | null)[]) => {
+		const node: ConcreteNode = nodeValue ? nodeValue.value : null;
+
+		if (node && node.firstChild) {
```
JosVerburg: I believe we should use the `executionParameters.domFacade` to retrieve the child node.
Note: `executionParameters` is now `_executionParameters`

PR #82, opened 2019-02-14
Path func

mehmetscoskun:


src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -112,6 +112,84 @@ const fnHasChildren: FunctionDefinitionType = function(
...
+const fnPath: FunctionDefinitionType = function(
+	_dynamicContext,
+	executionParameters,
+	_staticContext,
+	nodeSequence: ISequence
+) {
+	return zipSingleton([nodeSequence], ([nodeValue]: (Value | null)[]) => {
+		const node: ConcreteNode = nodeValue ? nodeValue.value : null;
+
+		if (!node) {
```
DrRataplan: I'd rather have an explicit null if here... And we can even move it to the first line of the function

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -112,6 +112,84 @@ const fnHasChildren: FunctionDefinitionType = function(
...
+		}
+
+		let result = '';
+
+		function getChildIndex(child: ConcreteChildNode): number {
+			let i = 0;
+			const childNodeType = child.nodeType;
+			const childNodeName = child.nodeName;
+			while (child !== null) {
+				if (childNodeType === child.nodeType && childNodeName === child.nodeName) {
```
DrRataplan: I'm missing tests for duplicate namespace URIs. I hoped the QT3 tesst offered them, but they don't...

Could you add the following tests:

```
let $dom := <xml>
  <a:a xmlns:a="http://test"/>
  <a:a xmlns:a="http://test2"/>
</xml>
return path(/xml/*:a[2])

(: Should output "fn:root()/Q{http://test2}a[1]" :)
```

```
let $dom := <xml>
  <a:a xmlns:a="http://test"/>
  <b:a xmlns:b="http://test"/>
</xml>
return path(/xml/*:a[2])

(: Should output "fn:root()/Q{http://test}a[2]" :)
```

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -112,6 +112,84 @@ const fnHasChildren: FunctionDefinitionType = function(
...
+		}
+
+		for (
+			let ancestor: ConcreteNode = node;
+			ancestor;
+			ancestor = executionParameters.domFacade.getParentNode(
+				ancestor as ConcreteChildNode
+			) as ConcreteNode
+		) {
+			if (
```
DrRataplan: Could you make this the 'exit condition' of the loop? like:

```
for (let ancestor: ConcreteNode = node;
domFacade.getParentNode(ancestor) !== null;
ancestor = ...) {
```

And move the returning part to after the loop?

This way the loop gets more readable~

test/assets/unrunnableTestCases.csv
```diff
@@ -842,6 +842,8 @@ fn-parse-json-944,AssertionError: expected [Function] to throw error matching /F
 fn-parse-json-945,AssertionError: expected [Function] to throw error matching /XPTY0004/ but got 'XPST0017: Function parse-json with arity of 2 not registered. Did you mean "Q{http://www.w3.org/2005/xpath-functions}parse-json (xs:stringnull)"?'
 fn-parse-json-946,Error: XPST0017: Function parse-json with arity of 2 not registered. Did you mean "Q{http://www.w3.org/2005/xpath-functions}parse-json (xs:stringnull)"?
 fn-parse-json-953,AssertionError: expected [Function] to throw error matching /FOJS0001/ but got 'XPST0017: Function codepoints-to-string with arity of 1 not registered. No similar functions found.'
+path014,SyntaxError: Expected "!", "!=", "*", "+", ",", "-", "/", "//", "<", "<<", "<=", "=", "=>", ">", ">=", ">>", "[", "and", "cast", "castable", "div", "eq", "except", "ge", "gt", "idiv", "instance", "intersect", "is", "le", "lt", "mod", "ne", "or", "to", "treat", "union", "|", "||", or "}" but "(" found.
+path017,Error: No selector counterpart for: computedTextConstructor.
 K-ContextPositionFunc-20,Error: No selector counterpart for: treatExpr.
```
DrRataplan: These two new flailing tests are to be expected, we don't do computed text constructors (yet) nor do we do whatever syntax construct is being used in the other test.

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -112,6 +112,84 @@ const fnHasChildren: FunctionDefinitionType = function(
 	});
 };
 
+const fnPath: FunctionDefinitionType = function(
```
JosVerburg: Could you also change this function to an arrow expression as that is preferred in our tslint configuration.

src/expressions/functions/builtInFunctions.node.ts
mehmetscoskun: ok, I did it

mehmetscoskun: ok, dI did it now

src/expressions/functions/builtInFunctions.node.ts
mehmetscoskun: done

src/expressions/functions/builtInFunctions.node.ts
mehmetscoskun: done

src/expressions/functions/builtInFunctions.node.ts
mehmetscoskun: it is done

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -112,6 +113,92 @@ const fnHasChildren: FunctionDefinitionType = function(
...
+					)}]${result}`;
+					break;
+				case NODE_TYPES.COMMENT_NODE:
+					result = `/comment()[${getChildIndex(ancestor)}]${result}`;
+					break;
+			}
+		}
+		if (ancestor.nodeType === NODE_TYPES.DOCUMENT_NODE) {
+			result = result || '/';
+			return sequenceFactory.create(createAtomicValue(result || '/', 'xs:string'));
```
DrRataplan: You've got a duplicate string defaulting here, could you make this just one?

Line 194 is unneeeded

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -112,6 +113,92 @@ const fnHasChildren: FunctionDefinitionType = function(
...
+		) {
+			switch (ancestor.nodeType) {
+				case NODE_TYPES.ELEMENT_NODE:
+					result = `/Q{${ancestor.namespaceURI || ''}}${
+						ancestor.localName
+					}[${getChildIndex(ancestor)}]${result}`;
+					break;
+				case NODE_TYPES.ATTRIBUTE_NODE:
+					result =
+						'/@' +
```
DrRataplan: Could you make this a template string?

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -3,7 +3,8 @@ import {
 	ConcreteNode,
 	ConcreteParentNode,
 	ConcreteProcessingInstructionNode,
-	NODE_TYPES
+	NODE_TYPES,
+	ConcreteElementNode
```
DrRataplan: Could you fix the order of imports?

src/expressions/functions/builtInFunctions.node.ts
mehmetscoskun: I deleted it

src/expressions/functions/builtInFunctions.node.ts
mehmetscoskun: yepp

src/expressions/functions/builtInFunctions.node.ts
mehmetscoskun: now it is in order

PR #89, opened 2019-02-21
Clone node

mehmetscoskun:


.vscode/launch.json
```diff
@@ -29,6 +29,7 @@
 				"fs.js",
 				"loader.js"
 			],
+			"smartStep": false,
```
JosVerburg: Please add this comment at the row above:
// smartStep can be enabled again once https://github.com/Microsoft/vscode/issues/68616 is fixed. TLDR; smartStep causes parts of code to not be highlighted when debugging unit tests.

PR #91, opened 2019-02-22
Fn translate

mehmetscoskun:


src/expressions/functions/builtInFunctions.string.ts
```diff
@@ -405,6 +405,32 @@ const fnNormalizeSpace: FunctionDefinitionType = function(
...
 
+const fnTranslate: FunctionDefinitionType = (
+	_dynamicContext,
+	_executionParameters,
+	_staticContext,
+	arg1,
+	arg2,
+	arg3
+) => {
+	const arr1 = Array.from(arg1.first() ? arg1.first().value : '');
```
DrRataplan: Can we use `zipSingleton` for these three variables? This will make the code more future-proof for synchronous evaluation

src/expressions/functions/builtInFunctions.string.ts
```diff
@@ -405,6 +405,33 @@ const fnNormalizeSpace: FunctionDefinitionType = function(
 	return sequenceFactory.singleton(createAtomicValue(string.replace(/\s+/g, ' '), 'xs:string'));
 };
 
+const fnTranslate: FunctionDefinitionType = (
```
JosVerburg: Could you try to use more descriptive names for arguments and variables? You can often take the names used in the spec description for inspiration. e.g.
`arg1` = `argSequence`
`arg2` = `mapStringSequence`
`arg3` = `transStringSequence`

`str1` = `arg`
`str2` = `mapString`
`str3` = `transString`

`e` = `character` or `c` as `c` is a commonly used as short for `character`

PR #93, opened 2019-02-26
Re-use externalDomFacade for the DombackedDomFacade

DrRataplan:
And remove DomBackedDomFacade. This leaves us with less code and more
consistent behaviour


DrRataplan:
No, the problem was with the unwrap being a problem. The unwrap method is called to give us the 'original' domFacade we gave to an evaluateXPath call, when we transfer execution to a custom XPath function. This domFacade in the XPath function turned out to sometimes be an internal domFacade if the outer evaluateXPath call did not pass a domFacade, effectively leaking an internal domFacade to the outside. This internal domfacade was closure-compiled, so all of it's keys were obfuscated.

PR #105, opened 2019-04-24
Fix API-extractor errors and Upgrade all dependencies

DrRataplan:


src/index.ts
```diff
@@ -93,13 +99,21 @@ if (typeof window !== 'undefined') {
...
+	/** @suppress {deprecated} */
+	// @ts-ignore We still need to expose this deprecated API
 	window['precompileXPath'] = precompileXPath;
 	window['registerXQueryModule'] = registerXQueryModule;
 	window['registerCustomXPathFunction'] = registerCustomXPathFunction;
 }
 
 export {
 	IDomFacade,
+	Options,
```
DrRataplan: This is the biggest change: export all types that we refer to in the public API

src/getBucketsForNode.ts
```diff
@@ -7,21 +7,23 @@ import { NODE_TYPES } from './domFacade/ConcreteNode';
  *
  * The bucket for a selector can be retrieved using {@link getBucketForSelector}.
  *
+ * @public
+ *
  * @param node - The node which buckets should be retrieved
  */
 export default function getBucketsForNode(
 	// TODO: Use the internal Node type for this parameter when we create one
```
DrRataplan: This TODO still stands and will be fixed in @Martinn1996's PR

src/index.ts
```diff
@@ -49,8 +51,8 @@ function parseXPath(xpathString: string) {
 }
 
 /**
- *
- * @param xpathString The XPath for which a buckets hould be retrieved
+ * @public
+ * @param xpathString - The XPath for which a buckets hould be retrieved
```
JosVerburg: Could you also fix the sentence? should be `The XPath for which buckets should be retrieved`

PR #106, opened 2019-04-25
Do not traverse the descendants axis for outermost

JosVerburg:
Fixes JIRA issue [DEV-2802](https://fontoxml.atlassian.net/browse/DEV-2802).

I also updated the tslint and added a tslint for the tests, these were not needed but reduce the amount of tslint warnings. Please confirm you agree with the added rules.


DrRataplan:
LGTM :shipit: :+1:

PR #117, opened 2019-05-17
DEV-2958 dynamicDontext includes currentContext property

mehmetscoskun:


mehmetscoskun:
I will add a test case for currentContext on Monday

PR #121, opened 2019-05-21
Dev 2386 buckets in axis

mehmetscoskun:


src/expressions/util/createDescendantGenerator.ts
```diff
@@ -11,28 +11,30 @@ import { DONE_TOKEN, IterationHint, ready } from './iterators';
 
 function findDeepestLastDescendant(
 	node: ConcreteNode,
-	domFacade: IWrappingDomFacade
+	domFacade: IWrappingDomFacade,
+	bucket
```
JosVerburg: Please use typing (at least for new code) `bucket: string`

src/expressions/util/createDescendantGenerator.ts
```diff
@@ -11,28 +11,30 @@ import { DONE_TOKEN, IterationHint, ready } from './iterators';
...
 	}
 	return parentNode;
 }
 
 export default function createDescendantGenerator(
 	domFacade: IWrappingDomFacade,
 	node: ConcreteNode,
-	returnInReverse = false
+	returnInReverse = false,
+	bucket
```
JosVerburg: Please use typing (at least for new code) `bucket: string`

src/types/Types.ts
```diff
@@ -52,9 +52,9 @@ export type Document = Node & {
  */
 export type Element = Node & {
 	localName: string;
-	namespaceURI: string;
+	namespaceURI: string | null;
```
JosVerburg: This seems to be part of issue #122. Do we need this in this PR?
Let's either:
- Revert this change and fix it for `Attr` and `Element` in another PR.
- Also make the fix for `Attr` in this PR. (In this case we should also notice in the comment something like "This fixes #122")

test/specs/parsing/axes/AncestorAxis.tests.ts
```diff
@@ -4,7 +4,9 @@ import jsonMlMapper from 'test-helpers/jsonMlMapper';
 
 import {
 	evaluateXPathToNodes,
-	evaluateXPathToStrings
+	evaluateXPathToStrings,
+	IDomFacade,
```
JosVerburg: Named imports must be alphabetized.tslint(ordered-imports)

test/specs/parsing/axes/AncestorAxis.tests.ts
```diff
@@ -28,6 +30,26 @@ describe('ancestor', () => {
...
+		jsonMlMapper.parse([
+			'parentElement',
+			['childElement']
+		], documentNode);
+
+		const parentNode = documentNode.firstChild;
+		const childNode = documentNode.firstChild.firstChild;
+
+		const testDomFacade: IDomFacade = {
+			getParentNode: (node, bucket) => {
```
JosVerburg: `getParentNode: (node: slimdom.Node, bucket: string): slimdom.Node | null => {`

test/specs/parsing/axes/AncestorAxis.tests.ts
```diff
@@ -28,6 +30,26 @@ describe('ancestor', () => {
...
+				chai.assert.notEqual(bucket, null, 'There must be a bucket passed!');
```
JosVerburg: We can use a specific null check `chai.assert.isNotNull(bucket, 'There must be a bucket passed!');`

test/specs/parsing/axes/AncestorAxis.tests.ts
```diff
@@ -28,6 +30,26 @@ describe('ancestor', () => {
...
+				return getBucketsForNode(node.parentNode).includes(bucket) ? node.parentNode : null;
```
JosVerburg: Seems like a somewhat creative way to check the right buckets. Lets use `chai.assert.include(getBucketsForNode(node.parentNode), bucket);` and just return `null`.

test/specs/parsing/axes/AncestorAxis.tests.ts
```diff
@@ -28,6 +30,26 @@ describe('ancestor', () => {
...
+			}
+		} as any;
+
+		const results = evaluateXPathToNodes('ancestor::parentElement', childNode, testDomFacade);
```
JosVerburg: When we do the check I described in the comment above, there will be no need left to check the result of xpath.
So this would than just be `evaluateXPathToNodes('ancestor::parentElement', childNode, testDomFacade);`.

test/specs/parsing/axes/AncestorAxis.tests.ts
```diff
@@ -28,6 +30,26 @@ describe('ancestor', () => {
 		], documentNode);
 		chai.assert.deepEqual(evaluateXPathToStrings('ancestor-or-self::*/position()', documentNode.documentElement.firstChild), ['1', '2']);
 	});
+
+	it('passes buckets for ancestor', () => {
```
JosVerburg: Comments on this test file also apply on the others. I.e. typing, tslint, explicit tests.

test/specs/parsing/axes/AncestorAxis.tests.ts
mehmetscoskun: The bucket comes from the selector, that's why `getBucketsFromNode(node.parentNode)` might not be include the bucket. for example, when parentNode is document node.

src/types/Types.ts
DrRataplan: Let's fix the other one(s) too and just fix it at the same type

test/specs/parsing/axes/AncestorAxis.tests.ts
```diff
@@ -14,50 +17,89 @@ beforeEach(() => {
...
+	});
+
+	it('passes buckets for ancestor', () => {
+		jsonMlMapper.parse(['parentElement', ['childElement']], documentNode);
+
+		const childNode = documentNode.firstChild.firstChild;
+
+		const testDomFacade: IDomFacade = {
+			getParentNode: (node: slimdom.Node, bucket: string | null) => {
+				chai.assert.include(getBucketsForNode(node.parentNode), bucket);
```
DrRataplan: I agree with this test. I like how it does not explicitly depend on the format of a bucket, but just tests whether it matches.

test/specs/parsing/axes/DescendantAxis.tests.ts
```diff
@@ -14,44 +12,68 @@ beforeEach(() => {
...
+
+		const childNode = documentNode.firstChild.firstChild;
+
+		const testDomFacade: IDomFacade = {
+			getFirstChild: (node: slimdom.Node, bucket: string|null) => {
+				chai.assert.include(getBucketsForNode(node.childNodes[0]), bucket, 'It includes bucket');
+				return null;
+			},
+			getNextSibling: (node: slimdom.Node, bucket: string|null) => {
+				return null;
```
DrRataplan: Could you also add the assert here?

test/specs/parsing/axes/DescendantAxis.tests.ts
```diff
@@ -14,44 +12,68 @@ beforeEach(() => {
...
+
+	it('passes buckets for descendant', () => {
+		jsonMlMapper.parse(['parentElement', ['childElement']], documentNode);
+
+		const childNode = documentNode.firstChild.firstChild;
+
+		const testDomFacade: IDomFacade = {
+			getFirstChild: (node: slimdom.Node, bucket: string|null) => {
+				chai.assert.include(getBucketsForNode(node.childNodes[0]), bucket, 'It includes bucket');
+				return null;
```
DrRataplan: If you return a childnode here, we will also execute the nextsibling method

test/specs/parsing/axes/FollowingAxis.tests.ts
```diff
@@ -98,6 +104,34 @@ return map{
...
+
+		const firstChildNode = documentNode.firstChild.firstChild;
+
+		const testDomFacade: IDomFacade = {
+			getFirstChild: (node: slimdom.Node, bucket: string|null) => {
+				return null;
+			},
+			getNextSibling: (node: slimdom.Node, bucket: string|null) => {
+				chai.assert.include(getBucketsForNode(node.nextSibling), bucket, 'It includes bucket');
+				return null;
```
DrRataplan: Let's return something here, so that we can also assert the other two parts

test/specs/parsing/axes/FollowingSiblingAxis.tests.ts
```diff
@@ -13,21 +11,59 @@ beforeEach(() => {
...
+		);
+
+		const firstChildNode = documentNode.firstChild.firstChild;
+
+		const testDomFacade: IDomFacade = {
+			getFirstChild: (node: slimdom.Node, bucket: string|null) => {
+				return null;
+			},
+			getNextSibling: (node: slimdom.Node, bucket: string|null) => {
+				chai.assert.include(getBucketsForNode(node.nextSibling), bucket, 'It includes bucket');
```
DrRataplan: Same, return, assert other two parts

test/specs/parsing/axes/PrecedingAxis.tests.ts
```diff
@@ -122,6 +124,34 @@ return map{
...
+		const testDomFacade: IDomFacade = {
+			getLastChild: (node: slimdom.Node, bucket: string|null) => {
+				return null;
+			},
+			getParentNode: (node: slimdom.Node, bucket: string|null) => {
+				return null;
+			},
+			getPreviousSibling: (node: slimdom.Node, bucket: string|null) => {
+				chai.assert.include(getBucketsForNode(node.previousSibling), bucket, 'It includes bucket');
+				return null;
```
DrRataplan: same, return node, assert two more

test/specs/parsing/axes/PrecedingSibling.tests.ts
```diff
@@ -13,21 +11,59 @@ beforeEach(() => {
...
+
+		const testDomFacade: IDomFacade = {
+			getLastChild: (node: slimdom.Node, bucket: string|null) => {
+				return null;
+			},
+			getParentNode: (node: slimdom.Node, bucket: string|null) => {
+				return null;
+			},
+			getPreviousSibling: (node: slimdom.Node, bucket: string|null) => {
+				chai.assert.include(getBucketsForNode(node.previousSibling), bucket, 'It includes bucket');
```
DrRataplan: same. return. assert

test/specs/parsing/axes/DescendantAxis.tests.ts
```diff
@@ -14,44 +12,69 @@ beforeEach(() => {
...
+	});
+
+	it('passes buckets for descendant', () => {
+		jsonMlMapper.parse(['parentElement', ['childElement']], documentNode);
+
+		const childNode = documentNode.firstChild.firstChild;
+
+		const testDomFacade: IDomFacade = {
+			getFirstChild: (node: slimdom.Node, bucket: string|null) => {
+				chai.assert.isNotNull(bucket, 'It has bucket');
```
DrRataplan: We can make this test whether it passes the bucket for 'childElement' (`getBucketsForSelector('self::childElement')`)

src/domFacade/IDomFacade.ts
```diff
@@ -6,13 +6,13 @@ import { Attr, CharacterData, Element, Node } from '../types/Types';
  * @public
  */
 export default interface IDomFacade {
-	getAllAttributes(node: Element): Attr[];
+	getAllAttributes(node: Element, bucket: string | null): Attr[];
```
DrRataplan: We could add some documentation here:

```
/**
 * Get all attributes of this element
 * @param  node
 * @param  bucket  The bucket that matches the attribute that will be used. This bucket can be used to narrow down which attributes should be retrieved.
 */

File: src/domFacade/IDomFacade.ts [hunk 19]
```diff
@@ -6,13 +6,13 @@ import { Attr, CharacterData, Element, Node } from '../types/Types';
...
+	getChildNodes(node: Node, bucket: string | null): Node[];
 	getData(node: Attr | CharacterData): string;
-	getFirstChild(node: Node): Node | null;
-	getLastChild(node: Node): Node | null;
-	getNextSibling(node: Node): Node | null;
-	getParentNode(node: Node): Node | null;
-	getPreviousSibling(node: Node): Node | null;
+	getFirstChild(node: Node, bucket: string | null): Node | null;
+	getLastChild(node: Node, bucket: string | null): Node | null;
+	getNextSibling(node: Node, bucket: string | null): Node | null;
```
DrRataplan:
```
/**
 * Get the next sibling of this node
 * @param  node
 * @param  bucket  The bucket that matches the nextSibling that is requested. An implementation of IDomFacade is free to interpret this to skip returning nodes that do not match the bucket, or use this information to its advantage.
 */
```

PR #129, opened 2019-06-03
Make debug mode and disabling cache a separate thing

DrRataplan:
The better error traces are realy useful when debugging
XPaths. Disabling caching is useful when working on the parser, but
makes performance very bad.

By making them separate, debugging XPaths will be easier.

### DIFF
--- a/src/parsing/compiledExpressionCache.ts
+++ b/src/parsing/compiledExpressionCache.ts
@@ -21,17 +21,26 @@ class CacheEntry {
 			| { namespaceURI: any; prefix: string }[]
 			| { namespaceURI: string; prefix: string }[]
 	) {
-		this.referredNamespaces = referredNamespaces;
-		this.referredVariables = referredVariables;
 		this.compiledExpression = compiledExpression;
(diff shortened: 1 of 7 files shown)


DrRataplan:
I mistakingly seem to have omitted a file in my original PR, could you re-check?

src/parsing/compiledExpressionCache.ts
```diff
@@ -102,6 +106,7 @@ export function getStaticCompilationResultFromCache(
 
 	const cacheWithCorrectContext = cachesForLanguage.find(
 		cache =>
+			cache.debug === debug &&
```
JosVerburg: This check ensures that the static evaluation is performed again for the expression. However this will not add or remove the `StackTraceGenerator`s from the expression.

So lets say we execute the same expression twice both with `disableCache: false`. But the first with `debug: true` and the second with `debug: false`. In this case the second one will also have the `StackTraceGenerator`s.

Wouldn't it make more sense to treat the debug flag the same as the language?

PR #145, opened 2019-07-24
Function return conversion

### LINKED ISSUE #125: XQuery functions do not convert their return value to the desired return type
XQuery function should convert the return value of a function to the desired return type as stated in [3.1.5.2 Function Conversion Rules](https://www.w3.org/TR/xquery-31/#id-function-conversion-rules) of the XQuery spec.

This can be observed with the following XQuery module:

```
declare %public function local:test ($node as node()) as xs:string {
	if ($node instance of text()) then
    	$node
    else
    	string($node)
};

let $nodes := //tip | //tip/text()
return $nodes/local:test(.)
```

In goes a mixed sequence of `text()` and `element()` nodes. Only the element nodes are cast explicitly to an xs:string, text nodes are returned as is. Then, the "/" operator starts complaining:

```
Error: XPTY0018: The path operator should either return nodes or non-nodes. Mixed sequences are not allowed.
```

[Playground example](https://xpath.playground.fontoxml.com/?mode=1&variables=%7B%7D&xml=%3Cxml%3E%0A++%3Ctitle%3Expath.playground.fontoxml.com%3C%2Ftitle%3E%0A++%3Csummary%3EThis+is+a+learning+tool+for+XML%2C+XPath+and+XQuery.%3C%2Fsummary%3E%0A++%3Ctips%3E%0A++++%3Ctip+id%3D%27edit%27%3EYou+can+edit+everything+on+the+left%3C%2Ftip%3E%0A++++%3Ctip+id%3D%27examples%27%3EYou+can+access+more+examples+from+a+menu+in+the+top+right%3C%2Ftip%3E%0A++++%3Ctip+id%3D%27permalink%27%3EAnother+button+there+lets+you+share+your+test+using+an+URL%3C%2Ftip%3E%0A++%3C%2Ftips%3E%0A%3C%2Fxml%3E&xpath=declare+%25public+function+local%3Atest+%28%24node+as+node%28%29%29+as+xs%3Astring+%7B%0A%09if+%28%24node+instance+of+text%28%29%29+then%0A++++%09%24node%0A++++else%0A++++%09string%28%24node%29%0A%7D%3B%0A%0Alet+%24nodes+%3A%3D+%2F%2Ftip+%7C+%2F%2Ftip%2Ftext%28%29%0Areturn+%24nodes%2Flocal%3Atest%28.%29)

DrRataplan:
We should validate/convert the return value of a function. This does break the build because of some small mistakes we made while annotating XPath standard functions, like `fn:name()`.

This fixes #125


test/specs/parsing/functions/FunctionCall.tests.ts
```diff
@@ -15,19 +16,40 @@ let documentNode;
 beforeEach(() => {
 	documentNode = new slimdom.Document();
 });
-
+d;
```
JosVerburg: Accidental key press?

PR #150, opened 2019-08-09
Add position to rethrown errors

jarivanmenxel:
This should not break anything existing, while giving extra information about the position the error occurred at in the code.


src/evaluationUtils/PositionedError.ts
```diff
@@ -0,0 +1,19 @@
+import { SourceRange } from '../expressions/debug/StackTraceGenerator';
```
JosVerburg: NIT: We do tend to add in some white lines for readability. e.g. between the import and export and before the constructor.

PR #152, opened 2019-08-13
Implement the default function namespace declarations

JCEntrena:


src/expressions/StaticContext.ts
```diff
@@ -37,6 +37,7 @@ export type FunctionDefinition = {
  */
 export default class StaticContext {
 	public parentContext: IContext;
+	private _registeredDefaultFunctionNamespace: string = 'http://www.w3.org/2005/xpath-functions';
```
JosVerburg: It doesn't really matter but we could just make this one public instead of adding a separate getter and setter.
Your choice.

test/specs/parsing/xquery/DefaultFunctionNamespace.tests.ts
```diff
@@ -0,0 +1,38 @@
...
+				documentNode,
+				undefined,
+				{},
+				{ language: evaluateXPath.XQUERY_3_1_LANGUAGE }
+			),
+			Math.PI
+		);
+	});
+	it('Can create a default namespace and a function', () => {
+		chai.assert.equal(
```
JosVerburg: chai does have an `isTrue` function which we usually use instead of equaling to true.
https://www.chaijs.com/api/assert/#method_istrue

test/specs/parsing/xquery/DefaultFunctionNamespace.tests.ts
```diff
@@ -0,0 +1,38 @@
...
+				undefined,
+				{},
+				{ language: evaluateXPath.XQUERY_3_1_LANGUAGE }
+			),
+			true
+		);
+	});
+
+
+});
```
JosVerburg: NIT files should end with a new line

src/parsing/processProlog.ts
```diff
@@ -88,6 +88,12 @@ export default function processProlog(
 
 		staticContext.registerNamespace(prefix, namespaceURI);
 	});
+	astHelper.getChildren(prolog, 'defaultNamespaceDecl').forEach(defaultNamespaceDecl => {
+		const namespaceURI = astHelper.getTextContent(
+			astHelper.getFirstChild(defaultNamespaceDecl, 'uri')
+		);
+		staticContext.setDefaultFunctionNamespace(namespaceURI);
```
JosVerburg: Can we add a check for the disallowed namespaces first? You could pretty much copy that from the handling of the `namespaceDecl`.
[Spec](https://www.w3.org/TR/xquery-31/#doc-xquery31-DefaultNamespaceDecl) says those same namespaces should also raise XQST0070.

src/parsing/processProlog.ts
```diff
@@ -88,6 +88,12 @@ export default function processProlog(
 
 		staticContext.registerNamespace(prefix, namespaceURI);
 	});
+	astHelper.getChildren(prolog, 'defaultNamespaceDecl').forEach(defaultNamespaceDecl => {
```
JosVerburg: There may only be 1 default namespace declaration.
From spec: "A Prolog may contain at most one default function namespace declaration [err:XQST0066]."

src/parsing/processProlog.ts
```diff
@@ -89,6 +90,37 @@ export default function processProlog(
 		staticContext.registerNamespace(prefix, namespaceURI);
 	});
 
+	// Default function namespace declaration
+	const defaultNamespaceFunctionDecl = astHelper
+		.getChildren(prolog, 'defaultNamespaceDecl')
+		.filter(x => {
+			return (
+				astHelper.getTextContent(astHelper.getFirstChild(x, 'defaultNamespaceCategory')) ===
```
DrRataplan: Could you add a not implemented error right here if the defaultNamespaceCategory is NOT equal to function? We still do not support the `default element namespace` syntax. We now 'accept' it and behave incorrectly. I would rather have an explicit 'NOT IMPLEMENTED' error here.

Koltharius: done

PR #153, opened 2019-08-13
DEV-3952: Implement the variable declarations

crodriguezq:


.gitignore
```diff
@@ -8,4 +8,6 @@
 /dist
 /src/parsing/xPathParser.raw.ts
 /.nyc_output
-/tmp
\ No newline at end of file
+/tmp
+.idea
+.vscode
```
DrRataplan: I think .vscode should not be in the gitignore. We actually include the 'launch.json' file there. Also, I would rather that the .idea folder is also not ignored. If your prefered IDE generates these files, I think it is your responsibility to not include those files, either by creating your own global gitignore, or manually not checking them in.

DrRataplan:
If you could discard the changes to the `.gitignore`, you have my approval!

franciscovelez:
Finished the changes. You can review now

PR #156, opened 2019-08-23
Feature/typeswitch

JCEntrena:
Implementation of the typeswitch funtion without the inner variable declaration.


src/expressions/NamedFunctionRef.ts
```diff
@@ -57,7 +56,7 @@ class NamedFunctionRef extends Expression {
 		let namespaceURI = this._functionReference.namespaceURI;
 		if (!namespaceURI) {
 			if (!this._functionReference.prefix) {
-				namespaceURI = FUNCTIONS_NAMESPACE_URI;
```
DrRataplan: I think the rebase did not yet really come through

src/parsing/compileAstToExpression.ts
```diff
@@ -1196,6 +1200,58 @@ function transformExpression(ast, compilationOptions) {
...
+function typeswitchExpr(ast: IAST, compilationOptions: CompilationOptions) {
+	const argExpr = compile(
+		astHelper.getFirstChild(astHelper.getFirstChild(ast, 'argExpr'), '*'),
+		compilationOptions
+	);
+
+	const caseClause = astHelper.getChildren(ast, 'typeswitchExprCaseClause');
+
+	const caseClauseExpressions = caseClause.map(X => {
+		let sequenceTypesAstNodes;
```
DrRataplan: I would rename X to 'caseClauseExpression' here

src/parsing/compileAstToExpression.ts
```diff
@@ -1196,6 +1200,58 @@ function transformExpression(ast, compilationOptions) {
...
+		if (astHelper.getChildren(X, 'sequenceTypeUnion').length !== 0) {
```
DrRataplan: I would invert this if to make it more readable

src/parsing/processProlog.ts
```diff
@@ -148,7 +162,7 @@ export default function processProlog(
 			);
 
 		if (!declarationNamespaceURI) {
-			throw new Error('XQST0060: Functions declared in a module must reside in a namespace.');
+			throw errXQST0060();
```
DrRataplan: Nice!

src/parsing/xpath.pegjs
```diff
@@ -294,7 +294,7 @@ ModuleImport
  = "import" S "module"
    prefix:(S "namespace" S prefix:$NCName _ "=" {return prefix})?
   // TODO: The URILiteral should be preceded with whitespace if the no prefix is set
-   _ uri:URILiteral (S "at" URILiteral (_ "," _ URILiteral)*)?
+   _ uri:URILiteral (S "at" S URILiteral (_ "," _ URILiteral)*)?
```
DrRataplan: Ah, good one!

test/assets/unrunnableTestCases.csv
```diff
@@ -4775,42 +4749,42 @@ d1e69567,Error: XPST0017: Function parse-ietf-date with arity of 1 not registere
...
 XMark-Q3,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
 XMark-Q4,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
 XMark-Q5,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
```
DrRataplan: I expected more succeeding tests. 26 new success is less then I expected

DrRataplan: I enabled the TypeSwitch tests in another blacklist we have. This causes more 'new' tests to succeed by running 144 previously not running tests!

src/expressions/NamedFunctionRef.ts
JCEntrena: I don't understand what you mean. Why is this a problem? 

What we are doing here is that instead of using the constant, we use the variable, which is the to the same value at the constant if there is no "default function namespace" specified.

test/assets/unrunnableTestCases.csv
```diff
@@ -108,7 +108,7 @@ fn-avg-6,Error: FORG0006: items passed to fn:avg are not all numeric.
 fn-avg-10,Error: FORG0006: items passed to fn:avg are not all numeric.
 cbcl-avg-004,Error: FORG0006: items passed to fn:avg are not all numeric.
 cbcl-avg-006,Error: FORG0006: items passed to fn:avg are not all numeric.
-cbcl-avg-008,Error: XPST0003: Unable to parse XPath: "typeswitch (fn:avg((xs:float(1), xs:double(2), xs:float(3)))) case $x as xs:double return $x default return "FAIL"". Expected "!", "!=", "*", "+", ",", "-", "/", "//", "<", "<<", "<=", "=", "=>", ">", ">=", ">>", "[", "and", "cast", "castable", "div", "eq", "except", "ge", "gt", "idiv", "instance", "intersect", "is", "le", "lt", "mod", "ne", "or", "to", "treat", "union", "|", "||", or end of input but "(" found. typeswitch [Error is around here](fn:avg((xs:float(1), xs:double(2), xs:float(3)))) case $x as xs:double return $x default return "FAIL"
+cbcl-avg-008,Error: XPST0008, The variable x is not in scope.
```
JosVerburg: I understood from @DrRataplan that we have not yet implemented this, can we throw a not implemented error instead of claiming the script is wrong.

test/assets/unrunnableTestCases.csv
```diff
@@ -4079,6 +4070,7 @@ XQST0089,AssertionError: expected [Function] to throw an error
 XQST0093a,Error: Duplicate function registration
 XQST0125_1,AssertionError: expected [Function] to throw an error
 XQST0125_2,AssertionError: expected [Function] to throw an error
+typeswitch-in-xpath,AssertionError: expected [Function] to throw an error
```
JosVerburg: This error seems to be new, can we investigate this and fix if possible?

src/parsing/xpath.pegjs
```diff
@@ -530,6 +531,33 @@ quantifiedExprInClause
      ]
    }
 
+// 74 
+TypeswitchExpr
+ = "typeswitch" _ "(" expr:Expr ")" _ clauses:(c:CaseClause _ {return c})+ "default" S varName:("$" v:VarName S {return v})? "return" S predicateExpr:ExprSingle
```
JosVerburg: predicates are used for filtering, can we just use something like `expr` or even `defaultExpr` for the default return to prevent possible confusion later on.

src/expressions/xquery/TypeSwitchExpression.ts
```diff
@@ -0,0 +1,108 @@
...
+			}
+		);
+
+		this._amountOfCases = caseClauses.length;
+		this._typeTestsByCase = caseClauses.map(clause => clause.typeTests);
+	}
+
+	public performFunctionalEvaluation(
+		dynamicContext: DynamicContext,
+		_executionParameters: ExecutionParameters,
```
JosVerburg: NIT: `_` prefix is usually used for arguments which aren't used in the function

src/expressions/NamedFunctionRef.ts
DrRataplan: yeah, but this was part of the default function namespace PR, right?

src/expressions/StaticContext.ts
```diff
@@ -37,6 +37,7 @@ export type FunctionDefinition = {
  */
 export default class StaticContext {
 	public parentContext: IContext;
+	public registeredDefaultFunctionNamespace: string = 'http://www.w3.org/2005/xpath-functions';
```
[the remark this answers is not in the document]
DrRataplan: agreed, but this should've been part of the default function namespace Pr

test/assets/unrunnableTestCases.csv
DrRataplan: I think the test here tests whether you can use the typeswitch expression in XPath. You shouldn't be able to. Could you add an error to the compiler like how for example the element constructors throw errors in the compiler?

src/parsing/compileAstToExpression.ts
```diff
@@ -1196,6 +1200,58 @@ function transformExpression(ast, compilationOptions) {
 	return new TransformExpression(transformCopies, modifyExpr, returnExpr);
 }
 
+function typeswitchExpr(ast: IAST, compilationOptions: CompilationOptions) {
+	const argExpr = compile(
```
DrRataplan: Could you add something similar as line 1097? an error that is thrown when we are not in 'XQuery' mode?

src/expressions/StaticContext.ts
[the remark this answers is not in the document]
JosVerburg: Might as well take it along in this PR as we now see it. Otherwise it'll likely be forgotten.
It would also be fine if we fix this directly on master as it should indeed have been fixed in another PR but was overlooked there.

test/assets/unrunnableTestCases.csv
JCEntrena: Here, we have to check is the declared variable is being used on the return clause of every case inside the typeswitch, and I don't now if there's an easy way to do that. Any suggestions? @JosVerburg @DrRataplan

test/assets/unrunnableTestCases.csv
JCEntrena: The tests tries to execute a typeswitch in xPath and expects an error, but adding the `!compilationOptions.allowXQuery` validation is not working, as it's never satisfied, even in that test. I don't know if there's another way to do that, but right now, that error is the expected behaviour.

src/parsing/compileAstToExpression.ts
JCEntrena: Sure, but as I've said in a comment above, it is not working on that test (and in general, it's never executed).

JCEntrena: The coveralls tests are failing because of this, as the condition is never satisfied. We can just revert the merge if you consider it appropriate.

test/assets/unrunnableTestCases.csv
JosVerburg: Yes, we checked and found the issue. To run the QT3 tests we [determine the language](https://github.com/FontoXML/fontoxpath/blob/master/test/qt3tests.ts#L472) based on the test file / test case.
A debug session indeed shows we use XQuery for this test case instead of XPath. The issue is that the language to use is either a descendant of the parent or of descendant of the test case.

We should always look at the descendants of the test case first before checking the descendants of the parent.

Could you fix that in this PR or make an issue for this so we can fix it later? Either way is ok with me, just would like to see this fixed ;)

JosVerburg: I did check this with the debug page and it does indeed throw the error which is correct behaviour.

PR #159, opened 2019-08-28
DEV-4033 Implement CompTextConstructor 

crodriguezq:


src/expressions/xquery/CompTextConstructor.ts
```diff
@@ -0,0 +1,36 @@
...
+	constructor(expr: Expression | null) {
+		super(expr ? expr.specificity : new Specificity({}), expr ? [expr] : [], {
+			canBeStaticallyEvaluated: false,
+			resultOrder: RESULT_ORDERINGS.UNSORTED
+		});
+
+		this._expr = expr;
+	}
+
+	public evaluate(_dynamicContext, executionParameters) {
```
JosVerburg: Lets add typing for these. `public evaluate(_dynamicContext: DynamicContext, executionParameters: ExecutionParameters) {`

JosVerburg: `_dynamicContext` is used so it shouldn't have the `_` prefix.

src/parsing/compileAstToExpression.ts
```diff
@@ -1064,6 +1067,17 @@ function computedCommentConstructor(ast, compilationOptions) {
 	return new CommentConstructor(expr);
 }
 
+function computedTextConstructor(ast, compilationOptions) {
```
JosVerburg: No need to type everything is this file, but I think we should make sure everything new is typed.
`function computedTextConstructor(ast: IAST, compilationOptions: CompilationOptions) {`

src/expressions/xquery/CompTextConstructor.ts
```diff
@@ -0,0 +1,36 @@
+import Expression, { RESULT_ORDERINGS } from '../Expression';
+import Specificity from '../Specificity';
+
+import castToType from '../dataTypes/castToType';
+import createNodeValue from '../dataTypes/createNodeValue';
+import sequenceFactory from '../dataTypes/sequenceFactory';
+
+class CompTextConstructor extends Expression {
```
JosVerburg: I suggest to remove the `Comp` prefix as we also do not use this for the other computed constructors.

crodriguezq:
Thanks for your comments @JosVerburg

PR #162, opened 2019-08-29
DEV-3957 Feature/flwor

Koltharius:


src/expressions/whereExpression.ts
```diff
@@ -0,0 +1,50 @@
...
+		let resultIterator: IAsyncIterator<Value> | null = null;
+		const whereExpressioResultSequence = sequenceCallbacks[0](dynamicContext);
+
+		return sequenceFactory.create({
+			next: (hint: IterationHint) => {
+				if (!resultIterator) {
+					const whereExpressionResult = whereExpressioResultSequence.tryGetEffectiveBooleanValue();
+
+					if (!whereExpressionResult.ready) {
+						return notReady(whereExpressionResult.promise);
```
Koltharius: The 0.002% coverage decrease is because we don't enter here in the test

JCEntrena: Removing this "if" doesn't break anything, so we can just delete it for good, if everybody agrees.

DrRataplan: it will if you put async expressions in the where, it will be hit:

```
for $i in (1,2,3)
where fontoxpath:sleep($i) > 1
return $i
```

We are not always testing this properly though, but in this case we can.

Koltharius: Ok, I'm going to add the test, thanks

test/assets/unrunnableTestCases.csv
```diff
@@ -2096,6 +2088,7 @@ K2-Axes-49,Error: No selector counterpart for: computedDocumentConstructor.
...
+K2-Axes-88,AssertionError: Expected XPath declare variable $i := <e> <a>1</a> <b>2</b> </e>; <result> { ($i/b, $i/a)/. } </result> to resolve to the given XML. Expected <result><b>2</b><a>1</a></result> to equal <result><a>1</a><b>2</b></result>
```
JosVerburg: #[Same as in this PR](https://github.com/FontoXML/fontoxpath/pull/159#discussion_r318608602)

src/parsing/compileAstToExpression.ts
```diff
@@ -407,52 +408,116 @@ function IfThenElseExpr(ast, compilationOptions) {
...
 	const returnClauseExpression = astHelper.getFirstChild(
 		intermediateClausesAndReturnClause[intermediateClausesAndReturnClause.length - 1],
 		'*'
 	);
 	const intermediateClauses = intermediateClausesAndReturnClause.slice(0, -1);
 
+	// We have to check if there are any intermediate clauses before compiling them.
 	if (intermediateClauses.length) {
 		if (!compilationOptions.allowXQuery) {
 			throw new Error('XPST0003: Use of XQuery FLWOR expressions in XPath is no allowed');
```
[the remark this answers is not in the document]
JCEntrena: Yeah, XPath only supports basic for/let expressions, so there cannot be any where/orderby/groupby expressions. Also, it only allows one for/let clause, which will be the initial clause. So we still need to check if there are more than one clause (that is, the presence of intermediate clauses), and throw the error if not using XQuery. 

We can check that on the [forClause XPath documentation](https://www.w3.org/TR/xpath-30/#doc-xpath30-SimpleForClause). It only admits one "for", and same holds for the let clause.

test/assets/unrunnableTestCases.csv
```diff
@@ -685,10 +680,10 @@ generate-id-016,Error: XPST0003: Unable to parse: "(:***************************
...
 generate-id-018,Error: XPST0017: Function generate-id with arity of 1 not registered. No similar functions found.
 generate-id-019,Error: XPST0017: Function generate-id with arity of 0 not registered. No similar functions found.
-generate-id-020,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
-generate-id-021,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
-generate-id-022,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
-generate-id-023,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
+generate-id-020,Error: XPST0017: Function generate-id with arity of 1 not registered. No similar functions found.
+generate-id-021,Error: XPST0017: Function generate-id with arity of 1 not registered. No similar functions found.
+generate-id-022,Error: No selector counterpart for: computedTextConstructor.
```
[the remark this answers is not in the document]
JCEntrena: Yeah, we'll have to do that after that branch is merged into master.

test/assets/unrunnableTestCases.csv
JCEntrena: This is weird. Just checked the result of the test on both `master` and this branch, and the result is the same. The expected result of the test is not correct.

test/assets/unrunnableTestCases.csv
```diff
@@ -1626,7 +1618,7 @@ K-NumericAdd-35,AssertionError: Expected executing the XPath "3 + (4, error())"
...
 K-NumericAdd-41,AssertionError: Expected executing the XPath "error() + 3" to resolve to one of the expected results, but got AssertionError: expected [Function] to throw error matching /FOER0000/ but got 'XPST0017: Function error with arity of 0 not registered. Did you mean "Q{http://www.w3.org/2001/XMLSchema}error (xs:anyAtomicType?)"?', AssertionError: expected [Function] to throw error matching /XPST0005/ but got 'XPST0017: Function error with arity of 0 not registered. Did you mean "Q{http://www.w3.org/2001/XMLSchema}error (xs:anyAtomicType?)"?'.
 K-NumericAdd-42,AssertionError: expected [Function] to throw error matching /FOER0000/ but got 'XPST0017: Function error with arity of 0 not registered. Did you mean "Q{http://www.w3.org/2001/XMLSchema}error (xs:anyAtomicType?)"?'
-op-numeric-add-1,Error: Not implemented: Intermediate clauses in flwor expressions are not implemented yet
+op-numeric-add-1,AssertionError:           for $x in (1, xs:decimal(2), xs:float(3), xs:double(4), xs:untypedAtomic(5))           for $y in (1, xs:decimal(2), xs:float(3), xs:double(4), xs:untypedAtomic(5))          return ($x + $y)       : expected false to be true
```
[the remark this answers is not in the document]
Koltharius: I checked it using the playground and the problem was that the assertion was wrong. [This](https://github.com/w3c/qt3tests/blob/23f29d53b0e71379d7008bb06397f1478f8579a4/op/numeric-add.xml#L1344) is the test. fontoxpath evaluates this to double instead of float.

PR #163, opened 2019-09-06
DEV-3958

JCEntrena:
Make the XQueryDoc implementation run in FontoXPath.


JCEntrena:
For some reason there is 1 test that fails on Travis. But it works on my machine (*insert "but it works on my machine memes"*). If someone could try to launch the qt3tests locally to double check, it would be nice.

src/expressions/functions/builtInFunctions.sequences.ts
```diff
@@ -289,6 +290,26 @@ const fnUnordered: FunctionDefinitionType = (
...
+const fnIndexOf: FunctionDefinitionType = (
+	dynamicContext,
+	_executionParameters,
+	_staticContext,
+	sequence,
+	search
+) => {
+	return search.mapAll(([onlySearchValue]) =>
+		sequence
+			.map((element, i) => {
```
JosVerburg: We miss atomization for the operands. See this example in the [specs](https://www.w3.org/TR/xpath-functions-31/#func-index-of): "The expression fn:index-of([1, [5, 6], [6, 7]], 6) returns (3, 4). (The array is atomized to a sequence of five integers)."

Regular [compares also do this](https://github.com/FontoXML/fontoxpath/blob/master/src/expressions/operators/compares/Compare.ts#L83) before calling `valueCompare`

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -98,6 +100,47 @@ const fnName: FunctionDefinitionType = function(
...
+		for (const value of allValues) {
+			if (isSubtypeOf(value.type, 'xs:anyAtomicType')) {
+				returnItems.push(value);
+				continue;
+			}
+			if (isSubtypeOf(value.type, 'node()')) {
+				returnItems.push(atomize(value, executionParameters));
+				continue;
+			}
+			if (isSubtypeOf(value.type, 'array()')) {
```
JosVerburg: This should be `array(*)` so it will accept arrays with any value.

test/assets/unrunnableTestCases.csv
```diff
@@ -428,6 +448,7 @@ fold-left-011,AssertionError: expected [Function] to throw an error
 fold-left-013,AssertionError: expected [Function] to throw error matching /XPTY0004/ but got 'Unrecognized type "elementTest".'
 fold-left-014,AssertionError: expected [Function] to throw error matching /XPTY0004/ but got 'Unrecognized type "elementTest".'
 fold-left-016,Error: XPST0008, The variable a is not in scope.
+fold-left-021,RangeError: Maximum call stack size exceeded
```
JosVerburg: Can we check what causes this? This seems rather dangerous if this is indeed a regression.

test/assets/unrunnableTestCases.csv
```diff
@@ -2656,77 +2636,64 @@ K2-DirectConElemWhitespace-5,Error: Not implemented: only module imports, namesp
...
 eqname-008,AssertionError: Expected XPath <out>{ (<a xmlns:ex="http://www.example.com/ns"><ex:b>93.7</ex:b></a>) /Q{http://www&#x2e;example&#x2E;com/ns}b }</out> to resolve to the given XML. Expected <out/> to equal <out><ex:b xmlns:ex="http://www.example.com/ns">93.7</ex:b></out>
-eqname-015,Error: Not implemented: references to variables with a namespace URI or a prefix.
 eqname-017,AssertionError: Expected XPath string-join(<a foo="3" bar="5" xml:space="preserve"/> / @Q{http://www.w3.org/XML/1998/namespace}*, '.') to resolve to "preserve": expected false to be true
-eqname-024,Error: Not implemented: references to variables with a namespace URI or a prefix.
+eqname-024,Error: XPST0008, The variable x is not in scope.
```
JosVerburg: This seems to indicate some parse errors concerning white space. (removing white spaces between foo and bar fixes prevents the error)

The test case:
```xml
<test-case name="eqname-024">
   <description>Tests that the whitespace is properly collapsed by comparing single-ws with multi-ws URIs</description>
   <created by="Abel Braaksma" on="2016-06-27"/>
   <dependency type="spec" value="XP30+ XQ30+"/>
   <test>for $Q{ urn:foo bar }x in 1 to 5 return $Q{urn:foo   bar}x + $Q{urn:foo bar}x</test>
   <result>
      <assert-string-value>2 4 6 8 10</assert-string-value>
   </result>
</test-case>```

File: test/assets/unrunnableTestCases.csv [hunk 5]
```diff
@@ -2656,77 +2636,64 @@ K2-DirectConElemWhitespace-5,Error: Not implemented: only module imports, namesp
...
-FunctionCall-014,Error: XPST0017: Function data with arity of 1 not registered. Did you mean "Q{http://www.w3.org/2001/XMLSchema}date (xs:anyAtomicType?)"?
+FunctionCall-014,AssertionError: Expected XPath          let $f := function($a as xs:integer, $b as xs:integer) as xs:integer {           data(<a>{$a}{$b}</a>)         }         return $f(12, 34)        to resolve to something of type xs:integer: expected false to be true
```
JosVerburg: I think the data function should cast the return type or the like? @DrRataplan could you clarify the approach for this, this seems to be the same issue you solved some time back (return types for functions).

test/specs/parsing/xquery/VariableDeclaration.tests.ts
```diff
@@ -10,97 +10,126 @@ beforeEach(() => {
...
+				{ language: evaluateXPath.XQUERY_3_1_LANGUAGE }
+			),
+			1
+		));
+
+	it('Test node operations and variable declaration scope - 2', () =>
+		chai.assert.equal(
+			evaluateXPathToNumber(
+				`
+					declare variable $x := 1;
```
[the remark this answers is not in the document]
JCEntrena: Oh, the test is duplicated on purpose. We had an error where the second (and ongoing) declarations of the same variable would not work. These paired tests aim to check if that works correctly or not. In the previous scenario, the first one would work, and the second would not. 

Should have noted that on a comment tho. Will do.

test/assets/unrunnableTestCases.csv
DrRataplan: yeah it should... I don't think this is related though

test/assets/unrunnableTestCases.csv
DrRataplan: Whitespace normalization in namespace uris is a whole can of worms we haven't even started to open, so I'm fine with leaving this as it is

src/expressions/functions/builtInFunctions.node.ts
```diff
@@ -98,6 +100,47 @@ const fnName: FunctionDefinitionType = function(
...
 
+const fnData: FunctionDefinitionType = function(
+	_dynamicContext,
+	executionParameters,
+	_staticContext,
+	sequence
+) {
+	function getDataFromValues(allValues: Value[]): Value[] {
+		let returnItems = [];
+		for (const value of allValues) {
```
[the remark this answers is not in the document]
DrRataplan: If you read it that way, we should always throw that error, none of or nodes have a typed value...

test/assets/unrunnableTestCases.csv
```diff
@@ -4399,7 +4387,7 @@ fo-test-fn-path-004,Error: No selector counterpart for: computedDocumentConstruc
 fo-test-fn-path-005,Error: No selector counterpart for: computedDocumentConstructor.
 fo-test-fn-path-006,Error: No selector counterpart for: computedDocumentConstructor.
 fo-test-fn-index-of-004,Error: XPTY0004: eqOp not available for xs:date and xs:integer
-fo-test-fn-index-of-005,Error: XPTY0004: eqOp not available for array(*) and xs:integer
+fo-test-fn-index-of-005,Error: Atomizing array(*) is not implemented.
```
Koltharius: I'm not sure about this error

src/expressions/functions/builtInFunctions.sequences.ts
```diff
@@ -299,6 +300,7 @@ const fnIndexOf: FunctionDefinitionType = (
 ) => {
 	return search.mapAll(([onlySearchValue]) =>
 		sequence
+			.atomize(_executionParameters)
```
Koltharius: Is this ok?

DrRataplan: That's ok, but usually we only use the underscore-prefixed variable names for unused variables. Since you're using it, I'd remove the underscore.

Koltharius: ok, I'm going to change it!

test/assets/unrunnableTestCases.csv
DrRataplan: That's a known one. We can ignore it.

PR #164, opened 2019-09-09
Bugfix/#11 custom xpath functions javascript parameter types

### LINKED ISSUE #155: Custom XPath functions that take in an xs:date or xs:dateTime do not get given a Javascript Date
Example function:
```
registerCustomXPathFunction(
            'fn:days-from-date',
            ['xs:date'],
            'xs:integer',
            (context: any, date: Date) => {
                return Math.floor(date.getTime() / 8.64e7)
            })
```

> TypeError: date.getTime is not a function

It returns this instead:
`w {
        h: 1970,
        a: 1,
        j: 1,
        o: 0,
        V: 0,
        C: 0,
        sa: 0,
        O: null,
        type: 'xs:date' }`

ymor:
Fixes #155
I hope you don't mind, but i also brought in a union type for the xml/xpath/xquery data types.  I've not updated all places where it could be used and are probably missing a few types.  I only covered just enough to get typescript to build.

### DIFF
--- a/src/evaluateXPath.ts
+++ b/src/evaluateXPath.ts
@@ -1,4 +1,3 @@
-import ExternalDomFacade from './domFacade/ExternalDomFacade';
 import IDomFacade from './domFacade/IDomFacade';
 import buildContext from './evaluationUtils/buildContext';
 import { printAndRethrowError } from './evaluationUtils/printAndRethrowError';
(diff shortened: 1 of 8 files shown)


ymor:
Sorry forgot to squash my commits

PR #165, opened 2019-09-09
Pass xml date types as javascript dates to custom functions

### LINKED ISSUE #155: Custom XPath functions that take in an xs:date or xs:dateTime do not get given a Javascript Date

ymor:
Fixes #155
I hope you don't mind, but i also brought in a union type for the xml/xpath/xquery data types. I've not updated all places where it could be used and are probably missing a few types. I only covered just enough to get typescript to build.


DrRataplan:
Great work, thanks! The union type makes sense too, for the xml schema related types!

I'll review the PR tomorrow, @JosVerburg could you also take a look?

src/expressions/dataTypes/Value.ts
```diff
@@ -1,3 +1,42 @@
 export default class Value {
-	constructor(public type: string, readonly value: any) {}
+	constructor(public type: ValueType, readonly value: any) {}
 }
+
+export type ValueType =
```
JosVerburg: Very nice!
I think we can even split these up in atomic specific and all types, or does that cause more issues? e.g.
```js
export type AnyAtomicType =
	| 'xs:boolean'
	| 'xs:string' ...;
export type ValueType =
	AnyAtomicType
	| 'node()'
	| 'element()'
	| 'document()'
	| 'text'
	| 'comment()'
	| 'item()'
	| 'function(*)'
	| 'map(*)'
	| 'array(*)';
```

ymor: You certainly can, but I'm not sure on the rules :) 
Would the idea be that the AtomicValue type could only take an AnyAtomicType? And Value would still take ValueType?

JosVerburg: Yes, that is indeed my idea. That way we prevent non atomic values ending up in an `AtomicValue`.
It doesn't have to be part of this PR btw, we can of course do further improvements like those sometime later.

ymor: Cool, well i had a quick go and it does cascade/get out of hand pretty quickly! So probably one for later :)

DrRataplan: Cool, something for later it is!

DrRataplan:
Thanks for your contribution! We will release this later this week, when the other big PR has come through

PR #174, opened 2019-10-16
Make external function declarations work correctly

DrRataplan:


Koltharius:
Solved, bad order in imports

DrRataplan:
@Koltharius thanks for fixing the imports!

PR #175, opened 2019-10-17
Updating functions

DrRataplan:
A lot of work went into this one. A quick summary:

- We are now able to evaluate an update script synchronously, to a predetermined type
- We are now able to evaluate calls to updating functions, which can be signaled using `%updating` or `public`
- We are now able to evaluate XQuery Update Script scripts as normal XQuery scripts, as long as we do not actually output a PUL. The `copy modify` language construct is one such example.

TODO: updating functions in XQuery modules. I would like you guys to have a chance in review while I fix that though!

Be gentle!


src/expressions/functions/FunctionCall.ts
```diff
@@ -54,80 +141,155 @@ class FunctionCall extends PossiblyUpdatingExpression {
...
-					if (functionItem.getArity() !== this._callArity) {
-						throw new Error(
-							`XPTY0004: expected arity of function ${functionItem.getName()} to be ${
-								this._callArity
-							}, got function with arity of ${functionItem.getArity()}`
-						);
+					const functionItem = validateFunctionItem(item, this._callArity);
+					if (functionItem.isUpdating) {
+						// dynamic function invocations cal not be updating
+						throw errXUDY0038();
```
JosVerburg: Missing import for this.

src/expressions/xquery-update/UpdatingExpression.ts
```diff
@@ -16,39 +47,11 @@ abstract class UpdatingExpression extends Expression {
 		this.isUpdating = true;
 	}
 
-	public ensureUpdateListWrapper(
```
JosVerburg: Any reason to move this function from the class?
I think we could have made it protected instead of public.

PR #176, opened 2019-10-22
Throw proper errors when using updating expressions in non-updating context

DrRataplan:


src/expressions/Expression.ts
```diff
@@ -26,6 +27,7 @@ abstract class Expression {
 	public peer: boolean;
 	public specificity: Specificity;
 	public subtree: boolean;
+	private _canBeUpdating: boolean = false;
```
bwrrp: This is always assigned in the constructor, the default here is not necessary

src/expressions/functions/FunctionCall.ts
```diff
@@ -17,7 +17,7 @@ import { errXPTY0004 } from '../XPathErrors';
 import { IPendingUpdate } from '../xquery-update/IPendingUpdate';
 import { mergeUpdates } from '../xquery-update/pulRoutines';
 import UpdatingFunctionValue from '../xquery-update/UpdatingFunctionValue';
-import { errXUDY0038 } from '../xquery-update/XQueryUpdateFacilityErrors';
+import { errXUST0001, errXUDY0038 } from '../xquery-update/XQueryUpdateFacilityErrors';
```
bwrrp: unused import

PR #182, opened 2019-11-04
Make inserting attributes with 'colliding namespaces' work

DrRataplan:
Attributes without a prefix should be in the null namespace.

### DIFF
--- a/src/expressions/xquery-update/InsertExpression.ts
+++ b/src/expressions/xquery-update/InsertExpression.ts
@@ -1,8 +1,15 @@
+import { ConcreteChildNode, ConcreteNode } from '../../domFacade/ConcreteNode';
+import { Attr, Element } from '../../types/Types';
+import isSubTypeOf from '../dataTypes/isSubtypeOf';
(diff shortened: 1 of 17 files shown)


src/expressions/xquery-update/InsertExpression.ts
```diff
@@ -42,7 +41,30 @@ export enum TargetChoice {
 	INSERT_INTO_AS_LAST = 5
 }
 
-function buildPendingUpdates(targetChoice, target, parent, alist, clist) {
+function testNamespaceURIForAttribute(targetElement, attributeNode: Attr, namespaceBindings): void {
+	const prefix = attributeNode.prefix || '';
+
+	// b. No attribute node in $alist may have a QName whose implied namespace binding conflicts with a namespace binding in the "namespaces" property of $target [err:XUDY0023].
+	const boundNamespaceURI = prefix ? targetElement.lookupNamespaceURI(prefix) : null;
```
DrRataplan: This here is the actual change. The rest is typing, typing and more typing

PR #189, opened 2019-12-18
Implement fn:matches

DrRataplan:
This works by requiring xspattern in two different ways: in unit tsts from node_modules, and in a
build using an UMD, and exposing it on window. Two different tsconfigs make this complete.


DrRataplan:
Not to be integrated yet. There is a bug: `fn:matches('aba', 'b')` should return true, while it is not

src/expressions/functions/builtInFunctions.string.ts
```diff
@@ -792,6 +817,14 @@ export default {
 			localName: 'codepoint-equal',
 			namespaceURI: FUNCTIONS_NAMESPACE_URI,
 			returnType: 'xs:boolean?'
+		},
+
+		{
+			argumentTypes: ['xs:string?', 'xs:string'],
```
JosVerburg: Could we also add the 2nd variant (with flags)?

src/index.ts
```diff
@@ -1,4 +1,5 @@
 import IDocumentWriter from './documentWriter/IDocumentWriter';
+
```
JosVerburg: Accidental enter hit?

src/expressions/functions/builtInFunctions.string.ts
DrRataplan: not easily no, I'd like to delay the flags for now and focus on a first push for the 2-arity matches

PR #191, opened 2019-12-20
Get language of the current qt3 test first

JCEntrena:
If we cannot retrieve it, or if it doesn't match any of the options, then we get the parent's language, as it was done before.


JCEntrena:
I may be missing some XPath options, and the xPath expression to get the language can probably be improved. I'll be waiting for your comments on the PR

PR #199, opened 2020-01-10
Throw the correct error when importing the same module twice

### LINKED ISSUE #198: Duplicate module imports throw unexpected errors
The test:
```
	it.only('crashes when importing the same module twice', () => {
		registerXQueryModule(`
module namespace test = "http://www.example.org/mainmodules.tests#2";

declare %public function test:hello($a) {
   "Hello " || $a
};
`);

		const result = evaluateXPath(
			`
import module namespace test = "http://www.example.org/mainmodules.tests#2";
import module namespace test = "http://www.example.org/mainmodules.tests#2";

test:hello("World")
`,
			null,
			null,
			null,
			null,
			{ language: evaluateXPath.XQUERY_3_1_LANGUAGE }
		);

		chai.assert.equal(result, 'Hello World');
	});
```

Throws: 
```
  1) Main modules
       crashes when importing the same module twice:
     Error: Duplicate function registration
```

That can't be the correct error, it should have a proper code

DrRataplan:
this fixes #198


DrRataplan:
There is just one QT3 test that is succeeding now... Not much coverage, unfortunately

PR #203, opened 2020-01-28
Allow a namespace resolver to be used instead of a callback function

DrRataplan:
The object can increase performance if you run XPaths like `evaluateXPathToBoolean('self::p', node)`
extremely frequently. This tries to use cached expressions and buckets to short-cut executions, but
determining whether a selector is re-usable for the same namespace config slowed it down.

### DIFF
--- a/src/evaluateXPath.ts
+++ b/src/evaluateXPath.ts
@@ -11,16 +11,67 @@ import convertXDMReturnValue, { IReturnTypes, ReturnType } from './parsing/conve
 import { Node } from './types/Types';
 
 /**
+ * Describes additional options for evaluateXPath and related functions
+ *
  * @public
  */
 export type Options = {
+	/**
+	 * The 'context' of this expression. This can be used for example to pass information around
+	 * from the call to 'evaluateXPath' to external functions.
+	 *
+	 * Note: not the same as the 'contextItem' in evaluateXPath.
+	 *
+	 * Defaults to `undefined`
+	 */
 	currentContext?: any;
+	/**
+	 * Setting 'debug' to true introduces better stack traces at the cost of some performance. Set
+	 * this to `true` when debugging and to `false` when deploying this.
+	 *
+	 * Defaults to `false`
+	 */
 	debug?: boolean;
-	documentWriter?: IDocumentWriter;
+	/**
+	 * Disable the selector cache, causing the parser to be used for every new selector.
+	 *
+	 * @internal
+	 */
 	disableCache?: boolean;
+	/**
+	 * Used to move nodes around. This is normally used to create a DOM tree. Default to using
+	 * methods like 'appendChild' on the nodes that are to be moved around.
+	 *
+	 * Observe that this will only be used for _new_ subtrees created using node
+	 * constructors. Existing nodes can only be mutated using XQuery Update Facility when calling
+	 * the {@link executePendingUpdateList} function.
+	 */
+	documentWriter?: IDocumentWriter;
+	/**
+	 * The Language used in the expression. Defaults to {@link EvaluateXPath.XPATH_3_1_LANGUAGE}.
+	 */
 	language?: Language;
+	/**
+	 * Additional modules to import. Register modules with {@link registerXQueryModule}. Defaults to no
+	 * modules being imported.
+	 */
 	moduleImports?: { [s: string]: string };
-	namespaceResolver?: (s: string) => string | null;
+	/**
+	 * Provides a mapping from prefixes used in the expression to their namespace URIs.
+	 *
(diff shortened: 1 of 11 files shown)


src/evaluateXPath.ts
```diff
@@ -11,16 +11,61 @@ import convertXDMReturnValue, { IReturnTypes, ReturnType } from './parsing/conve
...
 	moduleImports?: { [s: string]: string };
-	namespaceResolver?: (s: string) => string | null;
+	/**
+	 * Provides a mapping from prefixes used in the expression to their namespace uris.
+	 *
+	 * Can be an Object or a function. This is used to determine whether a cached expression can be
+	 * used: the semantics of a selector like `self::p` change when using a different namespace
+	 * resolver. Take care to not change the object, it is assumed to be frozen.
+	 */
+	namespaceResolver?: { [s: string]: string | null } | ((s: string) => string | null);
```
JosVerburg: This is currently the only option for which we do not describe the default value.

JosVerburg: Not necessarily better but consider this alternative text, it's how I would describe it  ;)

```
	 * Can be an object or a function. When an object is provided it may never change as cached
	 * expressions will be used if possible. For example the semantics of a selector like `self::p`
	 * change when using a different namespace resolver. When a function is provided all used
	 * prefixes will be tested before using cached expressions.
```

src/evaluateXPath.ts
```diff
@@ -11,16 +11,61 @@ import convertXDMReturnValue, { IReturnTypes, ReturnType } from './parsing/conve
...
+	/**
+	 * Used to create new elements. Defaults to the ownerDocument of the passed context node if absent
```
JosVerburg: NIT: "Used to create new elements" -> "Used to create new nodes".

src/expressions/ExecutionSpecificStaticContext.ts
```diff
@@ -60,24 +59,30 @@ export default class ExecutionSpecificStaticContext implements IContext {
...
+	public getReferredNamespaces(): { namespaceURI: string; prefix: string }[] | null {
 		return Object.values(this._referredNamespaceByName);
 	}
 
-	public getReferredVariables(): { name: string }[] {
-		return Object.values(this._referredVariableByName);
+	public getReferredVariables(): { name: string }[] | null {
+		const values = Object.values(this._referredVariableByName);
+		if (values.length === 0) {
+			return null;
```
JosVerburg: Why do we return null instead of the empty array? The array is already allocated so I don't see why we should not use it.

src/parsing/compiledExpressionCache.ts
```diff
@@ -186,6 +211,9 @@ export function storeStaticCompilationResultInCache(
 	cachesForLanguage.push(
 		new CacheEntry(
 			executionStaticContext.getReferredNamespaces(),
+			typeof namespaceResolver === 'function'
+				? null
```
JosVerburg: Should we not store the namespace resolver in the cache either way? It seems to me the `areNamespaceResolversEqual` will also handle functions which I don't will work if we don't save the function.

src/evaluateXPath.ts
```diff
@@ -11,16 +11,61 @@ import convertXDMReturnValue, { IReturnTypes, ReturnType } from './parsing/conve
...
 	currentContext?: any;
+	/**
+	 * Setting 'debug' to true introduces better stack traces at the cost of some performance. Set
+	 * this to `true` when debugging and to `false` when deploying this.
+	 *
+	 * Defaults to `false`
+	 */
 	debug?: boolean;
+	/**
+	 * Used to move nodes around. This is normally used to create a DOM tree. Default to using
```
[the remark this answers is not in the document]
DrRataplan: For creating them correct, nodesfactory, but for populating, we also use this one.

The expression `<ele><prrt /></ele>` will use the writer to append the `<prrt />` element to the `<ele />` element

src/expressions/ExecutionSpecificStaticContext.ts
DrRataplan: mainly to prevent the `.every` on an empty array later on. If the expression does not refer to variables, determining a cache hit is only a single null compare in this case

src/parsing/compiledExpressionCache.ts
DrRataplan: The function will not be used later on: we only reference compare the object variant.

For the function variant, we always use referred namespaces and test compatibility with the _new_ namespace resolver function.

DrRataplan:
Closed because the performance impact was negligible. It even caused a regression.

PR #208, opened 2020-02-03
Dev 5329 output context in x query traces used in hooks

Martinn1996:
Added an option `logOutput` `<function(string):void>` to output additional console statement when using the `trace` function. The readme has been updated accordingly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,8 @@ evaluateXPathToStrings(xpathExpression, contextNode, domFacade, variables, optio
     * `evaluateXPath.XQUERY_3_1_LANGUAGE` Evaluate `xpathExpression` according the [XQuery spec](https://www.w3.org/TR/xquery-31/).
   * `moduleImports` `<Object<string, string>`
   * `debug` `<boolean>` If a debug trace should be tracked, see [debugging](#debugging) for more information.
-  * `logOutput` `<function(string):void>` Log additional console statements for the trace function.
+  * `logger` `<Object>` Object with functions used to override the standard logger.
+    * `trace: <function(string):void>` The logger for the `trace()` function. The argument is the string of the original message.
 
 ### Example
 
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ evaluateXPathToStrings(xpathExpression, contextNode, domFacade, variables, optio
     * `evaluateXPath.XQUERY_3_1_LANGUAGE` Evaluate `xpathExpression` according the [XQuery spec](https://www.w3.org/TR/xquery-31/).
   * `moduleImports` `<Object<string, string>`
   * `debug` `<boolean>` If a debug trace should be tracked, see [debugging](#debugging) for more information.
+  * `logOutput` `<function(string):void>` Log additional console statements for the trace function.
 
 ### Example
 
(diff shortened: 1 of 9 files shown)


demo/main.ts
```diff
@@ -115,7 +115,10 @@ function jsonXmlReplacer(_key: string, value: any): any {
 async function runUpdatingXQuery(script: string) {
 	const result = await fontoxpath.evaluateUpdatingExpression(script, xmlDoc, null, null, {
 		debug: true,
-		disableCache: true
+		disableCache: true,
+		logOutput: m => {
```
DrRataplan: Let's add a `<pre>` element to the demo page and output the log to there, that we, we even have a good example use case

README.md
```diff
@@ -51,6 +51,7 @@ evaluateXPathToStrings(xpathExpression, contextNode, domFacade, variables, optio
     * `evaluateXPath.XQUERY_3_1_LANGUAGE` Evaluate `xpathExpression` according the [XQuery spec](https://www.w3.org/TR/xquery-31/).
   * `moduleImports` `<Object<string, string>`
   * `debug` `<boolean>` If a debug trace should be tracked, see [debugging](#debugging) for more information.
+  * `logOutput` `<function(string):void>` Log additional console statements for the trace function.
```
JosVerburg: This is not for logging additional console statements, this completely overrides the trace logs.
We should also indicate that the passed argument will be the original message generated by trace.

src/evaluateXPath.ts
```diff
@@ -16,9 +16,10 @@ import { Node } from './types/Types';
 export type Options = {
 	currentContext?: any;
 	debug?: boolean;
-	documentWriter?: IDocumentWriter;
 	disableCache?: boolean;
+	documentWriter?: IDocumentWriter;
 	language?: Language;
+	logOutput?: (message: string) => void;
```
JosVerburg: Have we really considered the name of this option?
All other logging e.g. errors when using `debug: true` will still be logged in the console and raised as error.

I propose to either rename this to `traceOutput` or also use it for errors for which we'd also need an indicator for the error.

Martinn1996: The options has been changed to an object `logger` which can contain multiple functions to override the standard logger. As for now, you can only override the logger for the `trace()` function.

demo/main.ts
```diff
@@ -115,7 +116,12 @@ function jsonXmlReplacer(_key: string, value: any): any {
 async function runUpdatingXQuery(script: string) {
 	const result = await fontoxpath.evaluateUpdatingExpression(script, xmlDoc, null, null, {
 		debug: true,
-		disableCache: true
+		disableCache: true,
+		logger: {
+			trace: m => {
+				traceOutput.innerHTML = m;
```
DrRataplan: I would append here, not overwrite.

demo/main.ts
```diff
@@ -130,7 +136,13 @@ async function runNormalXPath(script: string, asXQuery: boolean) {
 		disableCache: true,
 		language: asXQuery
 			? fontoxpath.evaluateXPath.XQUERY_3_1_LANGUAGE
-			: fontoxpath.evaluateXPath.XPATH_3_1_LANGUAGE
+			: fontoxpath.evaluateXPath.XPATH_3_1_LANGUAGE,
+		logger: {
+			trace: m => {
+				traceOutput.innerHTML = m;
```
DrRataplan: Same here

demo/main.ts
DrRataplan: Also, let's add a console.log here

bwrrp: consider using textContent instead of innerHTML to avoid having to escape things that look like HTML

DrRataplan:
lgtm. :shipit:

PR #212, opened 2020-02-10
DEV-5321 Implement lightweight cloning

mehmetscoskun:


src/expressions/ExecutionParameters.ts
```diff
@@ -1,12 +1,15 @@
...
 export default class ExecutionParameters {
 	constructor(
-		public readonly domFacade: IWrappingDomFacade,
+		public readonly domFacade: DomFacade,
 		public readonly nodesFactory: INodesFactory,
 		public readonly documentWriter: IDocumentWriter,
-		public readonly currentContext: any
+		public readonly currentContext: any,
+
+		public readonly realizedPointerByPointerMap: Map<NodePointer, NodePointer>
```
DrRataplan: suggestion: rootPointerByDescendantPointerMap?

src/domFacade/DomFacade.ts
```diff
@@ -1,135 +1,485 @@
...
+					))
+		) {
+			unwrappedParentNode = graftAncestor.parent as TinyParentNode | ConcreteParentNode;
+			newGraftAncestor = graftAncestor.graftAncestor;
+			unwrappedGrandParentNode = newGraftAncestor ? newGraftAncestor.parent : null;
+		} else {
+			// if not go to the dom
+			if (pointer.isTinyNode()) {
+				// if you have a tinyNode, it should not be possible, throw error
+				throw new Error('Boom, there should not be tinyNode at this point');
```
DrRataplan: let's remove these before merging please

src/domFacade/DomFacade.ts
```diff
@@ -1,135 +1,485 @@
...
 	ConcreteChildNode,
-	ConcreteElementNode,
-	ConcreteNode,
 	ConcreteParentNode,
 	NODE_TYPES
 } from './ConcreteNode';
 import IDomFacade from './IDomFacade';
-import IWrappingDomFacade from './IWrappingDomFacade';
+
+const isTinyNode = node => node.isTinyNode;
```
JosVerburg: Why do we need this as a separate function? I don't see the need to change a one liner to a one liner.

src/domFacade/DomFacade.ts
```diff
@@ -1,135 +1,485 @@
+import {
```
JosVerburg: Can we try to reduce the amount of casts in this file? There currently seem to 48 casts happening here.

I got the feeling that we either should take another look at the types we are using or rewrite it a bit so we don't have cast this often.

src/domFacade/DomFacade.ts
JosVerburg: Do we have the need for the error i.e. have we seen bugs which encounter this? If so, consider the error message "Boom" does not provide much information as it's already a raised `Error`

src/expressions/dataTypes/atomize.ts
```diff
@@ -31,50 +42,58 @@ export function atomizeSingleValue(
...
-				createAtomicValue(domFacade.getData(pointer), 'xs:untypedAtomic')
+				createAtomicValue(domFacade.getDataFromPointer(pointer), 'xs:untypedAtomic')
 			);
 		}
 
 		// comments and PIs are string
 		if (
-			isSubtypeOf(value.type, 'comment()') ||
-			isSubtypeOf(value.type, 'processing-instruction()')
+			pointer.node.nodeType === NODE_TYPES.COMMENT_NODE ||
```
JosVerburg, line 61: Is it no longer possible to have a custom type which inherits comments or processing instructions?
@DrRataplan I'm not very familiar with the custom types do you know if we can safely substitute with `isSubtypeOf` with an `===`?
Same goes for the other places in this function, several usages of `isSubtypeOf` seem to have been replaced.

src/expressions/dataTypes/createPointerValue.ts
```diff
@@ -24,10 +24,10 @@ function getNodeSubType(pointer: NodePointer, domFacade: DomFacade) {
 }
 
 export default function createPointerValue(pointer: NodePointer, domFacade: DomFacade): Value {
-	if (nodeValueCache.has(pointer)) {
-		return nodeValueCache.get(pointer);
-	}
+	// if (nodeValueCache.has(pointer)) {
```
JosVerburg, line 27: No need to have these as comments

src/expressions/dataTypes/documentOrderUtils.ts
```diff
@@ -49,17 +54,34 @@ function findAllAncestors(domFacade: DomFacade, node: NodePointer): NodePointer[
...
+			null
+		)
 	) {
 		ancestors.unshift(ancestor);
 	}
 
 	return ancestors;
 }
 
+function findAllActualAncestors(domFacade: DomFacade, node: ConcreteNode): ConcreteNode[] {
```
JosVerburg, line 68: Please keep this `findAllAncestors` and the other `findAllAncestorPointers`.
None of the two provides fake ancestors.

src/expressions/dataTypes/atomize.ts
DrRataplan: in practice, the dom types are not subclassed. If we add support for that, we should go over all of the code; this just saves a bit of performance

test/qt3testsBenchmark.ts
```diff
@@ -18,7 +18,7 @@ const suite: Suite = new Suite();
 
 function addTests(functions) {
 	Object.keys(functions).forEach((name) => {
-		suite.add(name, functions[name]);
+		suite.add(name, functions[name], {});
```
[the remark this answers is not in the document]
mehmetscoskun: it is not there now

src/expressions/dataTypes/createPointerValue.ts
mehmetscoskun: done

src/expressions/dataTypes/documentOrderUtils.ts
mehmetscoskun: done

Issue #214, opened 2020-02-11
how to add functions?


JoernT:
First of all - fontoxpath is a great effort!

I'd really like to dive deeper and maybe even contribute one day. However documentation is a bit sparse. I'm seriously missing a bit more elaborated examples of e.g. how to add new functions. 

I just don't get it from only reading the sources. If you'd like to see contributions you have to help it i fear (being myself an open source developer for many years) - is there any place where the fontoxpath is discussed?

Thanks a lot for this wonderful tool.

DrRataplan:
Hi, Thanks!

I agree documentation is a bit sparse, leaning to non-existant. Sorry about that! FontoXPath started as a 20% project here at [Fonto](https://www.fontoxml.com/), and it is mainly developed over here, at Fonto. Most communication happens over here as well, using an internal board. For external communication with non-Fonto devs, we use GitHub issues. Would you have any advice on what we can do more?

To answer your question about adding functions, we usually start by going over already-implemented functions.
Let's say we're going to implement something like `fn:contains-token#2` (https://www.w3.org/TR/xpath-functions-31/#func-contains-token) (note that collations are not implemented yet, so ignore the 3 arguments version). 

We'll need to start by adding the function definition at the end of the [builtInFunctions.string.ts]( https://github.com/FontoXML/fontoxpath/blob/master/src/expressions/functions/builtInFunctions.string.ts) file.
The first challenge would be to extract the 'actual' values from the lazy-evaluation iterator/generator-based arguments you would be getting in the function you've just created. A common pattern in FontoXPath is a function `next()` that returns one of three values: `{done: false, value: {type: <type>, value: <actual value>}}`, `{done: true}`, or `{ready: false, promise: Promise<undefined>}`. The first two values are vanilla iterators, but the third is different. FontoXPath is working towards an async implementation to prepare for doing things like a step-through debugger that can pause on a certain 'step', so that introduces additional complexity. These iterators are usually contained in an `ISequence`.
To make these iterators and sequences easier to work with, we've built several higher-order functions, like `mapAll` (get ALL values from an iterator, call the callback, and return an iterator over the return of the callback. Or Filter, which filters using a callback. See the `ISequence` class for more.

For the contains-token function, I'd use `ISequence#mapAll` for the first argument (we're going to need all of them anyway). This accepts a callback in which we can return another sequence. In this callback, I'd return the result of the mapAll function over the second argument.

This should look something like this:

```
[...]

PR #216, opened 2020-02-12
Update README.md

DrRataplan:
Mention we appreciate github issues.

Also, remove some text about a precompile action that we deprecated ages ago.


DrRataplan:
Addresses part of #214

Issue #217, opened 2020-02-13
access the AST


JoernT:
Sorry about my humble question - new to fontoxpath and unfortuantely also not a typescript guy , so please forgive if that's an obvious one:

how can i access the AST for a given XPath expression? I see, that there are functions for that in the source but these do not seem to be present in the dist (fontoxpath.js)

Not sure what it needs to get to that functionality.

Thanks

JosVerburg:
Hey,

The public API of fontoxpath does indeed not contain any feature to get the AST of an expression.
We do print the AST in the "demo" page we use when working on fontoxpath.

You can see [here](https://github.com/FontoXML/fontoxpath/blob/master/demo/main.ts#L154) that we get the AST through an internal API (and than format it a bit). You can [run this page](https://github.com/FontoXML/fontoxpath/blob/master/CONTRIBUTING.md#debugging) locally yourself.

Do you need to AST for something specific? So far we've not seen any need to expose the AST through public API.

And feel free to ask if you got any more questions or need some more help :)

JoernT:
Thanks for the reply. Sure, usually you don't need access to the AST. 

I'm just starting off with fontoxpath and use it for implementing a forms framework. Here you have expressions like this that attach constraints to some XML nodes:

<bind ref="foo" required="../bar = 'baz'">

Of course i can just iterate all my bindings and (re)evaluate them. However for more complex cases this will become very expensive. To circumvent that I'd like to know which nodes my given expression depends on (the above bind depends on node 'bar' which is a sibling of 'foo') and build a dependency graph that can be recalculated when nodes change. This way i can only reevaluate changed nodes and their dependencies instead of all (which makes a huge difference).

So, it would be super useful for me to have a way to access it.

DrRataplan:
Hi, cool! I like what you are doing, reminds me a lot of something I presented at the XML Prague conference a couple of years ago. We were building a schematron engine that could do live updates without recomputing everything all the time. `Soft validation in an editor environment` in https://archive.xmlprague.cz/2017/files/xmlprague-2017-proceedings.pdf. We went for a dynamic dependency tracking approach instead of a static one; you could consider that?

To answer your question, I would not mind exposing this API, we are currently parsing XPaths to XQueryX (in JsonML) internally. It would be trivial to implement a public function that accepts the _query_ and a `documentWriter` and `nodesFactory` that can parse the query, convert the `JsonML` AST to an actual DOM and output that.

JoernT:
[...]

PR #219, opened 2020-02-19
Add a way to parse a script to XQueryX

DrRataplan:
TODO: fix tests. I tried to make the XQueryX tests use this new API instead of the internal one, but
I did not manage to fix that properly

See #217 for one of the use-cases


DrRataplan:
TODO: provide documentation on the level of support of the XQueryX output. We align with the XQueryX tests as much as we can. We try to make the same choices as we saw there (for example, the XQueryX use nested sequences in some cases). We can however not yet parse every little construct.

src/parseScript.ts
```diff
@@ -0,0 +1,161 @@
...
+ * @param options - Additional options for parsing. Can be used to switch between parsing XPath or
+ * XQuery update facility
+ * @param  nodesFactory - A NodesFactory will be used to create the DOM. This can be a reference to
+ * the document in which the XQueryX will be created
+ * @param  documentWriter - The documentWrite will be used to append children
+ */
+export default function parseScript<TElement extends Element>(
+	script: string,
+	options: Options,
+	nodesFactory: INodesFactory,
```
JosVerburg: I think we should look into a way to have the `nodesFactory` also default to the dom backed version. A `document` does not match the signature as it does not implement `createDocument`.

test/xQueryXUtils.ts
```diff
@@ -1,66 +1,78 @@
...
 import * as chai from 'chai';
-import { sync, slimdom } from 'slimdom-sax-parser';
+import {
+	evaluateXPath,
+	evaluateXPathToBoolean,
+	evaluateXPathToNodes,
+	parseScript,
+	Language
+} from 'fontoxpath';
+import { slimdom, sync } from 'slimdom-sax-parser';
```
JosVerburg: We should import `slimdom` directly instead of via `slimdom-sax-parser` this will add it typed.

test/xQueryXUtils.ts
```diff
@@ -85,10 +97,10 @@ export function buildTestCase(testCase, loadXQuery, loadXQueryX, skippableTests,
 				chai.assert.equal(
 					new slimdom.XMLSerializer()
 						.serializeToString(actual.documentElement)
-						.replace(/></g, '>\n<'),
+						.replace(/></g, '\n><'),
```
JosVerburg: Is this correct? I would not expect the line break to be before the `>`

test/xQueryXUtils.ts
DrRataplan: yup, but it will break other things, like using the 'proper' serializer in slimdom

src/parseScript.ts
DrRataplan: I made this argument INodesFactory|Document. Document is the correct interface, but INodesFactory is the name we use for the same concept elsewhere

DrRataplan:
> I'd like to also use this new API in `demo/main.ts`, this seems also like a good test case to see if the API is easy to use. I did a quick attempt and noticed due to this that I could not pass a document as `INodesFactory`

Fixed, I made the demo page work correctly

src/nodesFactory/ISimpleNodesFactory.ts
```diff
@@ -0,0 +1,21 @@
+import { Attr, CDATASection, Comment, Element, ProcessingInstruction, Text } from '../types/Types';
+
+/**
+ * Subset of the constructor methods present on Document. Can be used to create textnodes, elements,
+ * attributes etcetera.
```
JosVerburg: Can we just write down all nodes you can create instead of using `etcetera`? They are not that many so we might as well, e.g. `Subset of the constructor methods present on Document. Can be used to create attributes, CDATA sections, comments, elements, processing instructions, and text nodes.`

PR #223, opened 2020-03-04
Correctly cache similar selectors

### LINKED ISSUE #222: Incorrect result when running the same xpath against a similar document
See below example test. 

```javascript
const nameSpaceDocument = new slimdom.Document();
const namespaceElement = nameSpaceDocument.createElementNS("http://test.com", "name");
namespaceElement.textContent = 'name1'
nameSpaceDocument.appendChild(namespaceElement);

chai.assert.equal(evaluateXPathToString('/name/text()',	nameSpaceDocument), 'name1')

const slimdomDocument = new slimdom.Document();
const nameElement = slimdomDocument.createElement("name");
nameElement.textContent = 'name2'
slimdomDocument.appendChild(nameElement);

chai.assert.equal(evaluateXPathToString('/name/text()',	slimdomDocument), 'name2')
```

The first document includes a namespace on the name element. The second document does not.
The second assertion will fail, however if you change the order of the assertions it works as expected. If you change the second xpath to just be '/name' it also works as expected.
Looks to be a caching issue?

DrRataplan:
Do not read from the actual cache when only intending to read from the half-compiled cache

Fixes #222

### DIFF
--- a/src/parsing/compiledExpressionCache.ts
+++ b/src/parsing/compiledExpressionCache.ts
@@ -100,11 +100,9 @@ export function getStaticCompilationResultFromCache(
 	const languageKey = generateLanguageKey(language, debug);
 	const cachesForLanguage = cachesForExpression[languageKey];
 	if (!cachesForLanguage) {
-		const halfCompiledExpressionFromCache = getAnyStaticCompilationResultFromCache(
-			selectorString,
-			language,
-			debug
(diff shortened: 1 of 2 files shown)


JosVerburg:
Changes seem good, coverage is going down so you expect the added test only traverses into one of the 2 if statements.

PR #232, opened 2020-03-25
Fix argument type of fn:string-join#1

### LINKED ISSUE #228: fn:string-join does not atomize input
The query `fn:string-join(1 to 9)` throws the error
```
Error: XPTY0004 Unable to cast argument of type xs:integer to type xs:string while calling string-join
  at <functionCallExpr>:1:1 - 1:23
```

This query should instead result in `"123456789"`. This is the [first example in the spec](https://www.w3.org/TR/xpath-functions-31/#func-string-join).

DrRataplan:
Fixes #228

### DIFF
--- a/src/expressions/dataTypes/casting/castToStringLikeType.ts
+++ b/src/expressions/dataTypes/casting/castToStringLikeType.ts
@@ -1,6 +1,8 @@
+import QName from '../valueTypes/QName';
+
 export default function castToStringLikeType(
-	instanceOf: (string) => boolean
-): (Value) => { successful: true; value: any } | { error: Error; successful: false } {
+	instanceOf: (t: string) => boolean
+): (value: any) => { successful: true; value: any } | { error: Error; successful: false } {
 	if (instanceOf('xs:string') || instanceOf('xs:untypedAtomic')) {
 		return value => ({
 			successful: true,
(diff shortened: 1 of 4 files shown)


src/expressions/functions/builtInFunctions.string.ts
```diff
@@ -694,7 +694,7 @@ export default {
 		{
 			namespaceURI: FUNCTIONS_NAMESPACE_URI,
 			localName: 'string-join',
-			argumentTypes: ['xs:string*'],
+			argumentTypes: ['xs:anyAtomicType*'],
```
DrRataplan: this here

JosVerburg: Could you also move this function declaration so it's besides the `string-join` one with 2 arguments? Every other overload seems to be next to one another except this one.
(Or move the other one of course, doesn't really matter)

src/expressions/dataTypes/casting/castToStringLikeType.ts
```diff
@@ -14,10 +17,15 @@ export default function castToStringLikeType(
 		});
 	}
 	if (instanceOf('xs:QName')) {
-		return value => ({
-			successful: true,
-			value: value.prefix ? `${value.prefix}:${value.localName}` : value.localName
-		});
+		return value => {
+			const qnameValue = value as QName;
```
JosVerburg: I think it's better to declare the type of value instead of casting here, i.e. `return (value: QName) => {` and remove `qnameValue` as `value` will now be recognised as a `QName`

src/expressions/dataTypes/casting/castToStringLikeType.ts
```diff
@@ -1,6 +1,9 @@
+import Value from '../Value';
```
JosVerburg: Unused import

PR #241, opened 2020-04-29
DEV-6507 Increase coverage again

mehmetscoskun:


mehmetscoskun:
Coverage before lightweight-cloning integration: 94.39%
Coverage after lightweight-cloning integration: 93.85%
Coverage at the this branch: 94.63%

src/expressions/dataTypes/documentOrderUtils.ts
```diff
@@ -25,6 +25,7 @@ function compareSiblingElements(
 	node2: NodePointer
 ): number {
 	if (arePointersEqual(node1, node2)) {
+		/* istanbul ignore next */
```
JosVerburg, line 28: I talked about this with Mehmet and it seems this is dead code for now. I understood we want to keep these checks in here in case we want to re-use this function elsewhere which I'm ok with.

But if we do add code elsewhere which suddenly cause these lines to no longer be dead code than we'd also have to remove these ignores which I doubt will happen as it's hard to keep track of.

So I'm ok with having some dead code to make functions more future proof but than we should also accept that the lines are not covered for now.

JosVerburg: Or we could remove the dead code as we do for `compareElements` a bit below.

test/specs/parsing/evaluateXPath.tests.ts
```diff
@@ -230,6 +237,17 @@ describe('evaluateXPath', () => {
 				() => evaluateXPathToFirstNode('fontoxpath:sleep(())', documentNode, domFacade),
 				'can not be resolved synchronously'
 			));
+
+		// const documentFragment = documentNode.createDocumentFragment();
```
JosVerburg, line 241: Commented code

mehmetscoskun: done

src/domFacade/DomFacade.ts
```diff
@@ -86,7 +86,11 @@ class DomFacade {
 	public getAttribute(pointer: ElementNodePointer, attributeName: string): string {
 		const node = pointer.node;
 		if (isTinyNode(node)) {
+			// The lines can be hit when fn:id or fn:idref functions are run with a target node
+			// whose root node is created by the document constructor.
```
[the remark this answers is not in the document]
mehmetscoskun: Because the documentConstructor has not been implemented yet.

src/expressions/dataTypes/documentOrderUtils.ts
mehmetscoskun: I am going to remove the dead code. If we need it in the future, we can add it.

test/specs/parsing/functions/functions.tests.ts
```diff
@@ -309,6 +319,19 @@ describe('functions', () => {
 			]);
 		});
 
+		it('returns nothing if the second parameter is not a node', () => {
```
DrRataplan, line 322: Should be 'returns nothing if the context is not a node'

mehmetscoskun: https://www.w3.org/TR/xpath-functions-31/#func-id
https://www.w3.org/TR/xpath-functions-31/#func-idref
`If the context item is absentDM31, dynamic error [err:XPDY0002]XP31.`
`If the context item is not a node, type error [err:XPTY0004]XP31.`
In this case, are expectations not compatible with the specifications of the functions(id and id-ref)?

mehmetscoskun:
If the commit `DEV-6507 Fix id and idref functions, and add tests` requires more work, I can revert it and create an issue for `id` and `idref` functions.

src/domFacade/DomFacade.ts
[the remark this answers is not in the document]
JosVerburg: Can we than remove these ignores? Same reason as the other places, we should either not have dead code or accept we have it.

src/expressions/functions/builtInFunctions.identifiers.ts
```diff
@@ -31,9 +32,15 @@ const fnId: FunctionDefinitionType = function (
 	targetNodeSequence
 ) {
 	const targetNodeValue = targetNodeSequence.first();
+	if (!targetNodeValue) {
+		throw new Error('XPDY0002: context is absent, it needs to be present to use id function.');
```
JosVerburg, line 36: Can we also move this error the the `XPathErrors` with `XPDY0002: context is absent, it needs to be present to use ${feature}` as message. No need to update the other places where it's used but lets try to keep using this pattern for errors we add.

mehmetscoskun: ok

PR #253, opened 2020-05-29
Use a merge sort for sorting nodes as a stable sort

### LINKED ISSUE #252: Differing union order results between browsers with a custom function
Hi,
We have this concept of cross document querying.  I've created the following as an example: https://codesandbox.io/s/mystifying-lalande-bxh5j?file=/src/App.tsx

We recently came across a scenario where we needed to do a union across these documents but the results are returning in a different order between chrome and firefox.  Now I'm not sure on the rules for the return values order - I noticed in the same document its document order. But I would have thought it should be consistent between browsers.

It's calling the functions in the same order, just the results get re-ordered.
Thanks.

ymor:
Fixes #252 
I added a union benchmark but the timings were all over the place on my machine for both the old and new code.  But generally they were within the same ball park.

Also added some headless browser tests to test the original issue.  Hopefully it wont be too much of a pain to add these to your CI!

### DIFF
--- a/karma.conf.ts
+++ b/karma.conf.ts
@@ -1,0 +1,21 @@
+module.exports = function (config) {
+	config.set({
+		frameworks: ['mocha', 'chai', 'karma-typescript'],
+		files: ['src/**/*.ts', 'test/browsertests.ts'],
+		preprocessors: {
+			'src/**/*.ts': 'karma-typescript',
+			'test/browsertests.ts': 'karma-typescript',
+		},
+		reporters: ['progress'],
+		port: 9876, // karma web server port
+		colors: true,
+		logLevel: config.LOG_INFO,
+		browsers: ['ChromeHeadless', 'FirefoxHeadless'],
+		autoWatch: false,
+		singleRun: true,
+		concurrency: Infinity,
+		karmaTypescriptConfig: {
+			tsconfig: 'test/tsconfig.json',
+		},
+	});
+};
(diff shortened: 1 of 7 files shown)


DrRataplan:
I extended you test a bit to perform a sort over 1k items, with the following test:

```javascript

function setup() {
document = sync(`
    <xml>
        <title>xpath.playground.fontoxml.com</title>
        <summary>This is a learning tool for XML, XPath and XQuery.</summary>
        <tips>
		${new Array(1000).fill('<tip>with some content</tip>')}
        </tips>
    </xml>`);
}

benchmarkRunner.only.addBenchmark(
	'union',
	() => {
		evaluateXPath('//tip | reverse(//tip)', document, domFacade);
	},
	setup
);
```

with your changes:
```
union x 0.44 ops/sec ±3.05% (6 runs sampled)
```
with the old sort:
```
union x 0.15 ops/sec ±2.13% (5 runs sampled)
```

Meaning the new implementation is at least in the same ball-park as the old one, which is nice. The is a test where a lot of sorting is happening in what I think is the worst case for quicksort, but the best case for merge sort. At least we have comparable performance, that's the most important to me.

I'll do some code cleanup and integrate it. Thanks!

ymor:
I guess the better way to solve this would be to handle the tiebreaker a bit better in the actual comparison when two nodes are from different documents.  Just not sure what that tiebreaker could be other than potential alphanumerical sort on name?

DrRataplan:
I think there is no real 'best' tie breaker in this case. Take for example the case where you have two identical nodes in two identical documents. There is no 'stable' way to differentiate between them. That's why we have a tiebreaker that at least creates the same output within the same run. I think the manual sorting improves on that by making it at least stable across browsers.

I feel that this might change though in the future; the order we now have could change to the reverse order, but still the same in all browsers / environments.

I'd say that if someone really feels strongly about the order of results in separate documents, they could always implement their own sorting algorithm to fix up the very last ordering problems.

Nonetheless, I'll integrate the manual sorting for now. Thanks for the PR!

Issue #211, opened 2020-02-10
array:sort does not sort array with sequence members properly


mehmetscoskun:
array:sort([(1,0), (1,1), (0,1), (0,0)]) should return [(0,0), (0,1), (1,0), (1,1)] but does not.

DrRataplan:
Ay, too bad. Could you tell us what it does return and how you are running the XPath?

In the playground, I'm getting this: https://xpath.playground.fontoxml.com/?mode=0&variables=%7B%7D&xml=%3Cxml%2F%3E&xpath=array%3Asort%28%5B%281%2C0%29%2C+%281%2C1%29%2C+%280%2C1%29%2C+%280%2C0%29%5D%29+%3D%3E+array%3Afor-each%28string-join%28%3F%2C+%27%2C%27%29+%29&context=

in: `array:sort([(1,0), (1,1), (0,1), (0,0)]) => array:for-each(string-join(?, ',') )`
out: 
```
Array with 4 items:
"0,0"
"0,1"
"1,1"
"1,0"
```

mehmetscoskun:
`array:flatten(array:sort([(1,0), (1,1), (0,1), (0,0)]))` is also an option to log it.

PR #270, opened 2020-09-21
 Reimplement array-sort

Abbasi-Waqas:
#211 Reimplemented array-sort based on the [w3 specs](https://www.w3.org/TR/xpath-functions-31/#func-sort)

- Modified ts-lint to include ordered imports


DrRataplan:
Seems like there is one test failing for some reason:

```
AssertionError: Expected XPath function-lookup(fn:QName('http://www.w3.org/2005/xpath-functions/array', 'sort'), 1)(["Red", "green", "blUE", "PINK", "ORanGE"]) to (deep equally) resolve to ["ORanGE", "PINK", "Red", "blUE", "green"]:
expected false to be true
```

Could you take a look at that?

src/expressions/functions/builtInFunctions_arrays.ts
```diff
@@ -393,27 +394,95 @@ const arrayForEachPair: FunctionDefinitionType = (
...
 
+const isString = (type: string): boolean => {
+	return (
+		isSubtypeOf(type, 'xs:string') ||
+		isSubtypeOf(type, 'xs:anyURI') ||
+		isSubtypeOf(type, 'xs:untypedAtomic')
+	);
+};
+
+const stringCompare = (arg1: string, arg2: string) => {
```
DrRataplan, line 405: I think this is unused, not?

DrRataplan (changes requested):

src/expressions/functions/builtInFunctions_arrays.ts
```diff
@@ -393,26 +404,123 @@ const arrayForEachPair: FunctionDefinitionType = (
...
+		).next(IterationHint.NONE).value
+	) {
+		return deepLessThan(
+			dynamicContext,
+			executionParameters,
+			staticContext,
+			valuesA.slice(1),
+			valuesB.slice(1)
+		);
+	} else if (valuesA[0].value !== valuesA[0].value) {
```
mehmetscoskun, line 441: this is always false, right?

mehmetscoskun (commented):
The actual  issue was:
array:sort([(1,0), (1,1), (0,1), (0,0)]) should return [(0,0), (0,1), (1,0), (1,1)] but does not.
If this is not covered by the qt3 tests, can you add this test as unittest.

src/expressions/functions/builtInFunctions_arrays.ts
Abbasi-Waqas: @mehmetscoskun This is a check to see if a value is NaN. Since Nan !== Nan, when valueA is Nan, it returns true (based on w3specs)
`else if ($A[1] ne $A[1] (:that is, $A[1] is NaN:)) 
     then fn:true()`

Abbasi-Waqas: I believe that test-case is already covered by the qt3tests.

DrRataplan (approved):

mehmetscoskun (approved):

PR #279, opened 2020-10-22
Implement first version of profiler

DrRataplan:


src/performance.ts
```diff
@@ -0,0 +1,162 @@
...
+	}
+	performance.mark(xpath);
+}
+export function markXPathEnd(xpath: string) {
+	if (!profilingEnabled) {
+		return;
+	}
+	// Replace the mark with a measure of the time spent
+	performance.measure(`xpath: ${xpath}`, xpath, undefined);
+	performance.clearMarks(xpath);
```
JosVerburg, line 161: Will this break if I call the same xpath function recursively twice? For this I could register a custom xpath function which keeps executing itself until an external value changes.

JosVerburg (approved):
Looks nice. Might be worth to also add some of the explanation to the README.md to make the feature easier to find. Could put it under the ##debug chapter or so.

src/performance.ts
DrRataplan: Fixed by adding a counter

DrRataplan:
Add tools for performance profiling

See README on how to use it

PR #287, opened 2021-02-04
DEV-5274 Add a factory to convert JS values into specific XDM types

JosVerburg:


src/evaluateXPath.ts
```diff
@@ -148,7 +149,7 @@ const evaluateXPath = <TNode extends Node, TReturnType extends keyof IReturnType
 	selector: string,
 	contextItem?: any | null,
 	domFacade?: IDomFacade | null,
-	variables?: { [s: string]: any } | null,
+	variables?: { [s: string]: TypedExternalValue | ValidValueSequence } | null,
```
DrRataplan, line 152: Maybe 'UntypedExternalValue'? This is going to be public facing

src/evaluationUtils/buildContext.ts
```diff
@@ -48,7 +54,7 @@ export default function buildEvaluationContext(
 	expressionString: string,
 	contextItem: any,
```
DrRataplan, line 55: Let's consider making the 'ContextItem' also a `TypedExternalValue|UntypedExternalValue` Removes another `any`

src/expressions/adaptJavaScriptValueToXPathValue.ts
```diff
@@ -69,16 +71,37 @@ function adaptItemToXPathValue(value: any, domFacade: DomFacade): Value | null {
 					})
 			);
 	}
+	// This code will be reached if the passed value is not actually a ValidValue. This can happen
+	// as JavaScript does not check the types from TypeScript
 	throw new Error(`Value ${value} of type "${typeof value}" is not adaptable to an XPath value.`);
 }
 
+function checkNumericType(value: ValidValue, type: ValueType): value is number {
```
DrRataplan, line 79: I'll take a second to update TypeScript so that we can use the `asserts value is number` syntax this afternoon. I'd prefer the `asserts` syntax if we can use it

src/types/TypedValueFactory.ts
```diff
@@ -0,0 +1,52 @@
...
+	[IS_XPATH_VALUE_SYMBOL]: true;
+	convertedValue: Value[];
+};
+
+/**
+ * Generates a factory to create a @see TypedExternalValue for the type @see typeName.
+ *
+ * @param  typeName  The type into which to convert the values.
+ */
+export default function (typeName: string) {
```
DrRataplan, line 35: Could you give this function a name? Maybe name it the same as the export from index.ts. Also please rename the file

PR #295, opened 2021-02-18
Allow external function declarations to pre-define a custom function

bwrrp:
This allows referring to that function before it is actually registered.


bwrrp:
We'll try to update this one to add a unit test or two next Monday

DrRataplan (approved):

ThomasGravekamp (approved):

JosVerburg (approved):

PR #296, opened 2021-02-22
Fix weird error with daytimeDurations being formatted wrong

DrRataplan:
And enable the tests for implemented features


src/expressions/dataTypes/valueTypes/DayTimeDuration.ts
```diff
@@ -59,13 +59,23 @@ class DayTimeDuration extends AbstractDuration {
...
 		const seconds = Math.abs(this.getSeconds());
-		const stringValue =
-			`${days ? `${days}DT` : 'T'}` +
-			`${hours ? `${hours}H` : ''}` +
-			`${minutes ? `${minutes}M` : ''}` +
-			`${seconds ? `${seconds}S` : ''}`;
+		const dayPart = `${days ? `${days}D` : ''}`;
+		const timePart =
+			(hours ? `${hours}H` : '') +
+			(minutes ? `${minutes}M` : '') +
```
JosVerburg, line 65: If I have the time 15:00:05 than won't that become 15H05S?
Wouldn't it make more sense to also include the minutes in that case or it defined like that in the spec?

DrRataplan: Correct, as spec.

JosVerburg (approved):

PR #298, opened 2021-03-09
Fix qt3performance

jarivm:
npm run qt3performance failed, and this is the fix.

Thank you @mehmetscoskun for helping me out.

PR #299, opened 2021-03-10
Set the compile target of unit tests to something sensible

DrRataplan:
A better way to fix #298

### DIFF
--- a/test/tsconfig.json
+++ b/test/tsconfig.json
@@ -8,7 +8,8 @@
 		},
 		"strict": false,
 		"allowJs": true,
-		"types": ["mocha", "node"]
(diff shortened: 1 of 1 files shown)


test/tsconfig.json
```diff
@@ -8,7 +8,8 @@
 		},
 		"strict": false,
 		"allowJs": true,
-		"types": ["mocha", "node"]
+		"types": ["mocha", "node"],
+		"target": "es2020"
```
bwrrp, line 12: To prevent future surprises we should probably check that all of the features this enables are supported in the node versions we run these tests against.

bwrrp (approved):

PR #303, opened 2021-03-27
Check for symbol in a polyfill compliant way

### LINKED ISSUE #302: Return type cardinality error on custom function in IE11
Hi,
We've come across an issue when upgrading from 3.14.2 to 3.15.0 when using a custom function.

Here's a really contrived example that I've tried to make a simple as possible: https://github.com/ymor/fonto-ie11  

If you run it in chrome/firefox, you will see the react logo. However if you use IE11 you will get the following error:
`XPTY0004: Multiplicity of function return value of type item() for my-function is incorrect. Expected exactly one`
I had a play around with the cardinality of the return type and it appears that fontoxpath thinks that the custom function is returning an empty sequence when running in IE11.  My assumption is that it's to do with the `createTypedValueFactory` that was added in 3.15.0. 

It is worth noting it still errors in the latest release. I was just trying to narrow it down to the release where it broke.
Also I'm not sure what your official stance is on supporting IE.  But I would really appreciate if you could support it for a little while longer :) 

Let me know if you have any further questions or have issues getting the repo running.
Thanks.

ymor:
Fixes #302.

### DIFF
--- a/src/registerCustomXPathFunction.ts
+++ b/src/registerCustomXPathFunction.ts
@@ -175,7 +175,7 @@ export default function registerCustomXPathFunction(
 			throw new CustomXPathFunctionError(error, localName, namespaceURI);
 		}
 
-		if (jsResult && typeof jsResult === 'object' && IS_XPATH_VALUE_SYMBOL in jsResult) {
+		if (jsResult && typeof jsResult === 'object' && Object.getOwnPropertySymbols(jsResult).includes(IS_XPATH_VALUE_SYMBOL)) {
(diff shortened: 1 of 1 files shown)


ymor:
Not sure why the lint stage has blow up.  I ran prettier and it made no further changes :/

DrRataplan (approved):

DrRataplan:
Awesome, thanks for the PR! I'll make a new release this week!

ymor:
Thanks for sorting out the error. I'll take a look at my prettier setup again at some point!

PR #304, opened 2021-03-29
No more async

DrRataplan:
Remove all asynchronous things from FontoXPath. This makes the codebase way easier to handle and easier to explain to new developers


src/expressions/dataTypes/Value.ts
```diff
@@ -1,5 +1,8 @@
+// The actual type is type ValueValue = NodePointer | Function | string | number | boolean | QName | Duration | DateTime; but doing that gives us thousands of errors.
+type ValueValue = any;
```
JosVerburg, line 2: Than why do we add a type for it now? We could as well just leave it as any and add a todo / issue to fix this at some point.

JosVerburg (approved):

src/expressions/dataTypes/Value.ts
DrRataplan: Meh, I'm going to keep it

PR #325, opened 2021-04-16
Implement JavaScript code generation

jarivm:
Implement code generation to JavaScript for a subset of XPath 3.1 queries. Although the subset is small, it covers frequently used queries. 

Compiling queries to JavaScript shows a significant (2-7x) improvement in query performance. Try running the benchmarks to look for yourself.

Supported XPath 3.1 functionalities are:

- path expressions;
- self, parent, attribute and child axes;
- predicates;
- "and" and "or" operators; and
- a bunch of node tests, including name tests.

The implementation exposes two new functions: `compileXPathToJavaScript` and `executeJavaScriptCompiledXPath`. FontoXPath's users can decide if and how to use them. Both are marked experimental. Types associated to these functions are exposed as well.

Includes benchmarks, unit tests and coverage checking with the qt3tests.


src/evaluateXPath.ts
```diff
@@ -136,6 +147,112 @@ const evaluateXPath = <TNode extends Node, TReturnType extends keyof IReturnType
...
+			domFacade,
+			variables,
+			returnType,
+			options
+		);
+	} else if (backend === 'js-codegen') {
+		return evaluateWithJsCodegenBackend(selector, contextItem, domFacade, returnType, options);
+	} else if (backend === 'auto') {
+		try {
+			return evaluateWithJsCodegenBackend(
```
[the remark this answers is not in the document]
jarivm: > Hm, this try the code gen, if it breaks -> just use the expression backend feels silly.

I agree. In some cases errors are thrown because an AST can not be compiled because it is unsupported, while in other times it may throw for different reasons.

> Instead, how would you feel to expose a function that scans an AST for feasibility for codegen, but returns something that it does not work. Otherwise, if an expression breaks during execution, it will just try again in the Expression backend. Which feels silly.

I don't think this is the best option. If I understand correctly, this would mean two functions have to be maintained: one that scans the AST and one that transforms it into JS code. Since I think the js codegen backend should throw errors when it can't compile an AST, I feel like this is duplication of the same idea.

What if we make a distinction between errors? If the thrown error is of type `UnsupportedAST`, then try compiling the AST with the expression backend. If the error is of a different type, then it should be handled differently (maybe just rethrow the error, which is what happens when the expression backend encounters an error).

src/evaluateXPath.ts
```diff
@@ -136,6 +147,112 @@ const evaluateXPath = <TNode extends Node, TReturnType extends keyof IReturnType
...
+
+		compiledJavaScript = compileAstToJavaScript(queryBodyContents, returnType);
+		storeCompiledJavaScriptInCache(expressionString, returnType, compiledJavaScript);
+	}
+	const wrappedDomFacade: DomFacade = new DomFacade(
+		domFacade === null ? new ExternalDomFacade() : domFacade
+	);
+
+	return compiledJavaScript.evaluate(
+		{ type: 'document-node()', value: { node: contextItem } },
```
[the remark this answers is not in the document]
jarivm: Done

src/jsCodegen/compileAstToJavaScript.ts
```diff
@@ -0,0 +1,422 @@
+import { ValueType } from '../expressions/dataTypes/Value';
```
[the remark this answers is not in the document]
jarivm: Done

src/types/Options.ts
```diff
@@ -60,6 +60,24 @@ export type NamespaceResolver = (prefix: string) => string | null;
...
+	 * Defaults to the stable 'expression' backend.
+	 *
+	 * FontoXPath has two backends: the older 'expression' backend and the
+	 * experimental 'js-codegen' backend. The js-codegen backend can be a lot
+	 * faster than the expression backend. If you want to try it out, use the
+	 * 'auto' option. This ensures that, if the js-codegen fails to compile a
+	 * given query, the expression backend is used instead. WARNING: this is
+	 * experimental.
+	 *
+	 * The 'js-codegen' option is used for TESTING ONLY.
```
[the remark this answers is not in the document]
jarivm: Done, used @beta since @experimental was rejected by `npm run build`. 

> Warning: built/types/Options.d.ts:68:8 - (tsdoc-unsupported-tag) The TSDoc tag "@experimental" is not supported by this tool

src/evaluateXPath.ts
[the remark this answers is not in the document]
DrRataplan: Not when you make the API something like this:

```
function compileToJavaScript(ast: IAst): {astIsAcceptable: true, result: string?}|{astIsAcceptable: false}
```

That boolean can just be returned instead of an error.

Later check whether the AST is acceptable, in the codegen branch just throw if it isn't but in the automatic flow, just fall back to JavaScript.

This is also a bit more usable in the future when we pull this parse + compile to the build phase of the Fonto editor: we can just scan the JavaScript (or Typescript at that point) and replace any and all XPaths with the compiled JavaScript counterpart.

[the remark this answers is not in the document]
jarivm: Do you think it would be a good idea to specify a message with `{astIsAcceptable: false}`? Like `{astIsAcceptable: false, reason: string}`? This can make it easier to see why the AST is not accepted, which can make debugging easier. Maybe if the reason is supplied one can decide to implement the feature in the backend.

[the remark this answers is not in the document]
jarivm: Done

DrRataplan:
## BundleMon

  
  
  
  
  
<details open>
<summary>Files updated (2)</summary>

Status | Path | Size | Limits
:------------: | ------------ | :------------: | :------------:
:white_check_mark: | fontoxpath.js<br/> | 86.01KB (+2.32KB +2.77%) | -
:white_check_mark: | fontoxpath.esm.js<br/> | 86.1KB (+2.31KB +2.76%) | -
</details>

  
  


Total files change +4.63KB +2.76%

Final result: :white_check_mark:

---
<p align="center"><a href="https://bundlemon.now.sh/projects/609024effb87a20009a6b096/reports?branch=js-codegen&resolution=all" target="_blank" rel="noreferrer noopener">Current branch size history</a> | <a href="https://bundlemon.now.sh/projects/609024effb87a20009a6b096/reports?branch=master&resolution=days" target="_blank" rel="noreferrer noopener">Target branch size history</a></p>

src/evaluateXPath.ts
```diff
@@ -136,6 +149,101 @@ const evaluateXPath = <TNode extends Node, TReturnType extends keyof IReturnType
 		throw new TypeError("Failed to execute 'evaluateXPath': xpathExpression must be a string.");
 	}
 
+	const backend = options && options.backend ? options.backend : 'expression';
+
+	if (backend === 'js-codegen' || backend === 'auto') {
+		const compiledJavaScriptResult = compileWithJsCodegenBackend(selector, returnType, options);
+
+		if (compiledJavaScriptResult.isAstAccepted === true) {
```
DrRataplan, line 157: `if (a.b=== true)`  === `if (a.b)`

src/domFacade/ConcreteNode.ts
```diff
@@ -9,7 +9,7 @@ import {
 	Text,
 } from '../types/Types';
 
-export const enum NODE_TYPES {
+export enum NODE_TYPES {
```
DrRataplan, line 12: Why did you make this change? the `export const enum` causes it to be inlined, I think

src/evaluateXPath.ts
jarivm: Ah, TS garandeert natuurlijk dat het een boolean is.

jarivm: Done

src/evaluateXPath.ts
```diff
@@ -136,6 +149,101 @@ const evaluateXPath = <TNode extends Node, TReturnType extends keyof IReturnType
...
+		if (!mainModule) {
+			// This must be a library module
+			throw new Error('Can not execute a library module.');
+		}
+
+		const prolog = astHelper.getFirstChild(mainModule, 'prolog');
+		const queryBodyContents = astHelper.followPath(mainModule, ['queryBody', '*']);
+
+		if (prolog) {
+			throw new Error('Unsupported: XQuery.');
```
DrRataplan, line 225: Don't you want to just reject it in this case, instead of throwing?

src/evaluateXPath.ts
```diff
@@ -136,6 +149,101 @@ const evaluateXPath = <TNode extends Node, TReturnType extends keyof IReturnType
...
+		? null
+		: getCompiledJavaScriptFromCache(expressionString, returnType);
+
+	if (!compiledJavaScriptResult) {
+		const ast = parseExpression(expressionString, compilationOptions);
+		const mainModule = astHelper.getFirstChild(ast, 'mainModule');
+
+		if (!mainModule) {
+			// This must be a library module
+			throw new Error('Can not execute a library module.');
```
DrRataplan, line 218: Also here, don't you wanna reject instead of throwing? Now you're likely missing the special XPath error code.

src/jsCodegen/CompiledJavaScript.ts
```diff
@@ -0,0 +1,29 @@
...
+export type CompiledJavaScriptResult =
+	| { isAstAccepted: true; result: CompiledJavaScript }
+	| { isAstAccepted: false; reason: string };
+
+export default class CompiledJavaScript {
+	private compiledJavaScript: string;
+	private fn: any;
+	private runtimeLibrary: any;
+
+	constructor(compiledJavaScript: string, runtimeLibrary: any) {
```
DrRataplan, line 13: I am still not sure whether including this here is the best idea. This does make using the codegen backend very easy, but it also blocks off generating JavaScript files on disk that contain the precompiled XPaths.

The codegen back-end is very much in beta and it will be changed. I'm kind-of fine keeping this as is, but like I said, not sure.

@jarivm what are your thoughts?

DrRataplan: Also, lets type the runtimeLibrary

src/domFacade/ConcreteNode.ts
jarivm: Yes, I did it so generated code could use it from the runtime lib. However, after you mentioned it I went looking for another way to do it and found that inlining the enum value in the generated code also works (which is better). Fixed it.

src/jsCodegen/CompiledJavaScript.ts
jarivm: I've replaced it with this:

```
let compiledJavaScriptFunction = compilationOptions.disableCache
    ? null
    : getCompiledJavaScriptFromCache(expressionString, returnType);

if (!compiledJavaScriptFunction) {
    const ast = parseExpression(expressionString, compilationOptions);
    const compiledJavaScriptResult = compileAstToJavaScript(ast, returnType);

    if (compiledJavaScriptResult.isAstAccepted) {
        // tslint:disable-next-line
        compiledJavaScriptFunction = new Function(
            'contextItem',
            'domFacade',
            'runtimeLibrary',
            compiledJavaScriptResult.code
        );

        storeCompiledJavaScriptInCache(
            expressionString,
            returnType,
            compiledJavaScriptFunction
        );
    }
}

if (compiledJavaScriptFunction) 
    return evaluateCompiledJavaScript(compiledJavaScriptFunction, contextItem, domFacade);
} else {
    // other things
}
```

`evaluateCompiledJavaScript` and `compileAstToJavaScript` could be exposed. That way storing compiled JS is possible. 

We should, however, look for an alternative for the `new Function`. .

DrRataplan: The easiest way to prevent a `new Function` is I guess to follow your comment and expose a `compileXPathToJavaScript` function that accepts all the static info: namespace bindings, expected return type and the other static things, returning a JavaScript source code string. Plus an `executeCompiledXPath` function that accepts an instance of a function, a node, etcetera and evaluates the function with the runtime lib plus all the arguments.

That way, users can choose to either `new Function` the JS source code string to a function instance, or just output the acquired string to disk to load it later. When storing it on disk, I think it is likely they will be something like ``` const precompiledXPaths = {'self::p': function (runtimeLib, node, domFacade) {<whatever the compiler barfed out>}, 'self::div': <more compiled code>} . That code can be minified, served, parsed by the browser and passed into the evaluate function.

jarivm: Would this work?
```javascript
it.only('compiles XPath to an executable JavaScript function', () => {
  const supportedQuery = '/xml';
  const unsupportedQuery = '/xml[@id = "yes"]';

  // Compile time.
  const precompiledXPaths = {};
  function compileIfPossible(query: string, returnType: ReturnType) {
    const jsResult = compileXPathToJavaScript(query, returnType);
    if (jsResult.isAstAccepted) {
      precompiledXPaths[generateKey(supportedQuery, returnType)] = new Function(
        'contextItem',
        'domFacade',
        'runtimeLibrary',
        jsResult.code
      ) as () => {};
    }
  }

  compileIfPossible(supportedQuery, ReturnType.BOOLEAN);
  compileIfPossible(unsupportedQuery, ReturnType.BOOLEAN);

  // Run time.
  function executeQuery(query: string, returnType: ReturnType) {
    const precompiledQuery = precompiledXPaths[generateKey(query, returnType)];
    if (precompiledQuery) {
      chai.assert.isTrue(executeCompiledXPath(precompiledQuery, document));
    } else {
      // Use expression backend.
      chai.assert.isTrue(evaluateXPathToBoolean(query, document));
    }
  }

  executeQuery(supportedQuery, ReturnType.BOOLEAN);
  executeQuery(unsupportedQuery, ReturnType.BOOLEAN);
}
```
I see one issue with avoiding `new Function` in FontoXPath regarding the 'auto'/fallback option. Somehow FontoXPath has to be able to use `new Function` on its own when using 'auto'. Maybe the right way to do this would be to `fontoxpath.enableFunction(Function);`. The advantage of this would be that the user does not have to know about the arguments a executable XPath as JavaScript function can receive. The downside is that use of `new Function` is not transparent at all. This solution looks kind of weird as well.

Maybe, if there is no better way, the 'auto' and 'js-codegen' options should be removed and be replaced with the `compileXPathToJavaScript` and `executeCompiledXPath` API's where the user is responsible. That way there are no 'hidden' uses of `new Function` and is using js-codegen very much opt-in. The downside is that more responsibility is shifted to the user, who has to know more about FontoXPath's internals, which can restrict future API changes. This would, however, be a good thing for security, since all use of `new Function` is explicit. This does not take away the problem that generated JavaScript could contain malicious code which is evaluated anyway... 

What do you think?

test/specs/jsCodegen/astRejection.tests.ts
```diff
@@ -0,0 +1,191 @@
...
+
+declare %public function test:hello($a) {
+"Hello " || $a
+};
+`,
+				null,
+				null,
+				null,
+				{ backend: 'js-codegen' }
+			);
```
DrRataplan, line 188: I like to add explicit checks for the correct error type so that we can be sure we are not throwing a TypeError by accident because we made a typo in this error handling code

test/specs/jsCodegen/kindTests.tests.ts
```diff
@@ -0,0 +1,27 @@
+import * as chai from 'chai';
+import * as slimdom from 'slimdom';
+
+import jsonMlMapper from 'test-helpers/jsonMlMapper';
+
+import evaluateXPathToBoolean from '../../../src/evaluateXPathToBoolean';
```
DrRataplan, line 6: The TypeScript integration in VS code sometimes does this.... 

```
import {evaluateXPathToBoolean} from 'fontoxpath';
```

test/specs/jsCodegen/operators.tests.ts
```diff
@@ -0,0 +1,35 @@
+import * as chai from 'chai';
+import * as slimdom from 'slimdom';
+
+import jsonMlMapper from 'test-helpers/jsonMlMapper';
+
+import evaluateXPathToBoolean from '../../../src/evaluateXPathToBoolean';
```
DrRataplan, line 6: also check these imports

test/specs/jsCodegen/predicates.tests.ts
```diff
@@ -0,0 +1,67 @@
+import * as chai from 'chai';
+import * as slimdom from 'slimdom';
+
+import jsonMlMapper from 'test-helpers/jsonMlMapper';
+
+import evaluateXPathToBoolean from '../../../src/evaluateXPathToBoolean';
+import evaluateXPathToNodes from '../../../src/evaluateXPathToNodes';
```
DrRataplan, line 7: and these imports

src/jsCodegen/emitTest.ts
```diff
@@ -0,0 +1,92 @@
...
+	if (namespaceURI) {
+		return rejectAst('Unsupported: namespace URI in name tests.');
+	}
+
+	if (prefix === '*' && localName === '*') {
+		return acceptAst(conditionCode);
+	}
+
+	return acceptAst(
+		`${conditionCode} && ${identifier}.localName === ${JSON.stringify(localName)}`
```
DrRataplan, line 43: aren't we missing a Namespace test here?

`self::p` should only match the `p` element when it is in the namespace the namespaceResolver resolves the empty prefix to

test/specs/jsCodegen/evaluateXPathWithJsCodegen.ts
```diff
@@ -0,0 +1,51 @@
...
+	returnType?: ReturnType
+): IReturnTypes<TNode>[TReturnType] => {
+	returnType = returnType || (ReturnType.ANY as any);
+	query = normalizeEndOfLines(query);
+
+	const cachedQuery = cache[generateKey(query, returnType)];
+
+	if (!cachedQuery) {
+		const compiledXPathResult = compileXPathToJavaScript(query, returnType);
+		if (compiledXPathResult.isAstAccepted === true) {
```
DrRataplan, line 36: Let's see next Friday _why_ this is needed. There might be something in the typing of this object that we can improve.

src/index.ts
```diff
@@ -219,4 +226,9 @@ export {
 	profiler,
 	registerCustomXPathFunction,
 	registerXQueryModule,
+	compileXPathToJavaScript,
```
DrRataplan, line 229: Can you alfabetize this?

DrRataplan: Can you also export this on line 129? That's the place where we export fontoxpath for use directly in the browser with globals. The old-fashioned way.

src/jsCodegen/compileAstToJavaScript.ts
```diff
@@ -0,0 +1,127 @@
...
+	return (contextItem, domFacade, runtimeLibrary) => {
+		const {
+			DONE_TOKEN,
+			ValueType,
+			XPDY0002,
+			adaptSingleJavaScriptValue,
+			determinePredicateTruthValue,
+			isSubtypeOf,
+			ready,
+		} = runtimeLibrary;
```
DrRataplan, line 66: I am actually surprised this works in the closure-compiled version of fontoxpath. I expected all these names to be renamed to `a` or `$`. Can you verify this all works in the bundle as well?

DrRataplan: You can verify this by including the jsCodeGen tests in the `npm run integrationtests` script.

DrRataplan (approved):

PR #342, opened 2021-05-10
Type rework

HectorPeeters:
This pull requests converts the old string-based type system to a more robust enum based solution. All types are now represented as ValueTypes, SequenceTypes or ParameterTypes.

### DIFF
--- a/src/expressions/dataTypes/Value.ts
+++ b/src/expressions/dataTypes/Value.ts
@@ -251,11 +251,7 @@ const stringToValueTypeMap: { [key: string]: ValueType } = {
  * @returns A string corresponding to that base type.
  */
 export function valueTypeToString(input: ValueType): string {
-	const stringVal = typeToStringMap[input];
-	if (stringVal === undefined) {
-		throw new Error(`Cannot convert ValueType of type "${input}" to a string`);
-	}
-	return stringVal;
+	return typeToStringMap[input];
 }
 
 /**
--- a/src/expressions/dataTypes/isSubtypeOf.ts
+++ b/src/expressions/dataTypes/isSubtypeOf.ts
@@ -1,5 +1,5 @@
 import builtinDataTypesByType, { TypeModel } from './builtins/builtinDataTypesByType';
-import { startWithXS, ValueType } from './Value';
+import { ValueType } from './Value';
 import { Variety } from './Variety';
 
 function isSubtypeOfType(subType: TypeModel, superType: TypeModel) {


.gitlab/issue_templates/issue_template.md
```diff
@@ -0,0 +1,14 @@
+# Issue
```
DrRataplan, line 1: Lets remove this.

pnpm-lock.yaml
```diff
@@ -0,0 +1,4779 @@
+lockfileVersion: 5.3
```
DrRataplan, line 1: I guess you're Yarn users or something? Lets remove this artefact

src/registerCustomXPathFunction.ts
```diff
@@ -127,8 +128,8 @@ function splitFunctionName(
  */
 export default function registerCustomXPathFunction(
 	name: string | { localName: string; namespaceURI: string },
-	signature: string[],
-	returnType: string,
+	signature: SequenceType[],
```
DrRataplan, line 131: As discussed, these should remain of type `string`

test/specs/expressions/adaptJavaScriptValueToXPathValue.tests.ts
```diff
@@ -7,98 +8,129 @@ import { domFacade as adaptingDomFacade } from '../../../src';
...
 		chai.assert.equal(xpathSequence.first().value, 1, 'is 1');
 	});
 
-	it('does not support unknown types', () => {
-		chai.assert.throws(
-			() => adaptJavaScriptValueToSequence(null, 1, 'fonto:theBestType'),
-			' can not be adapted to equivalent XPath values'
-		);
-	});
+	// TODO: This is impossible right?
```
DrRataplan, line 20: That error is still to be expected. It's just that the type that will be pased in should be a string, not the enum value

src/index.ts
```diff
@@ -15,6 +15,16 @@ import evaluateXPathToNumbers from './evaluateXPathToNumbers';
 import evaluateXPathToString from './evaluateXPathToString';
 import evaluateXPathToStrings from './evaluateXPathToStrings';
 import executePendingUpdateList from './executePendingUpdateList';
+import castToType from './expressions/dataTypes/castToType';
+import createAtomicValue from './expressions/dataTypes/createAtomicValue';
+import {
```
DrRataplan, line 20: I would not expect _any_ changes in this file at all

src/expressions/dataTypes/Value.ts
```diff
@@ -5,67 +5,390 @@ export default class Value {
...
+]);
+
+/**
+ * Creates a string value from a base type, generally follows the 'xs:type' notation, where
+ * xs stands for XML Schema
+ * @param input The base type to get the string of
+ * @returns A string corresponding to that base type.
+ */
+export function valueTypeToString(input: ValueType): string {
+	const typeToStringMap: Record<ValueType, string> = {
```
DrRataplan, line 198: you can hoist this map to the module level. right now it looks like it needs to be declared whenever the function starts

src/expressions/dataTypes/Value.ts
```diff
@@ -5,67 +5,390 @@ export default class Value {
...
+
+	return valueTypeToString(input.type);
+}
+
+export function stringToValueType(input: string): ValueType {
+	if (!input.startsWith('xs:') && input.indexOf(':') >= 0) {
+		throw new Error('XPST0081: Invalid prefix for input ' + input);
+	}
+
+	const stringToValueTypeMap: { [key: string]: ValueType } = {
```
DrRataplan, line 296: this map can also be on the module level

src/expressions/dataTypes/Value.ts
```diff
@@ -5,67 +5,390 @@ export default class Value {
...
+		'comment()': ValueType.COMMENT,
+		'item()': ValueType.ITEM,
+		'function(*)': ValueType.FUNCTION,
+		'map(*)': ValueType.MAP,
+		'array(*)': ValueType.ARRAY,
+	};
+
+	const type = stringToValueTypeMap[input];
+
+	if (type === undefined || type === null) {
```
DrRataplan, line 364: only a check for `undefined` is plenty

src/expressions/dataTypes/builtins/dataTypeValidatorByType.ts
```diff
@@ -159,47 +159,41 @@ function isValidDayTimeDuration(value: string): boolean {
...
-
-	'xs:integer': isValidInteger,
-
-	'xs:yearMonthDuration': isValidYearMonthDuration,
-	'xs:dayTimeDuration': isValidDayTimeDuration,
-};
-
-export default dataTypeValidatorByName;
+	const validatorToType: Map<ValueType, (value: string) => boolean> = new Map([
+		[ValueType.XSANYSIMPLETYPE, isValidAnySimpleType],
```
DrRataplan, line 163: Lets also hoist this map to the module level. It is constant, no need to redeclare it

src/index.ts
HectorPeeters: Since some of the tests also use ValueType, EllipsisType, ... Can they stay in the export section or should all code that gets called by the tests get converted to strings?

test/specs/expressions/adaptJavaScriptValueToXPathValue.tests.ts
HectorPeeters: So should `adaptJavaScriptValueToSequence` take in a string instead of a SequenceType, or should we add stringToValueType to the test itself. Since that is now the function that throws an error if a type isn't found.

DrRataplan: Ah sure, this is one of the few actual unit tests that test internal code. Feel free to cover this in another way, such as (trying to register) a custom function that has a nonsense return type: `chai.assert.throws(()=>registerCustomXPathFunction('test:func', [], 'this-type::does-not-exist', ()=>{}), 'Whatever error it throws')`

src/index.ts
DrRataplan: Tests that use that type can just import them directly from the package. You'll see only unit tests in the `expressions` folder import it, not the integration tests in the `parsing` folder

src/expressions/dataTypes/Value.ts
```diff
@@ -5,67 +5,390 @@ export default class Value {
...
-	| 'processing-instruction()'
-	| 'comment()'
-	| 'item()'
-	| 'function(*)'
-	| 'map(*)'
-	| 'array(*)';
+/**
+ * The ValueTypes;
+ * Previously represented by strings.
+ * @public
```
DrRataplan, line 11: Nope, not public.

Maybe document it as ```The value types. Every value in XPath is of one of these types``` or something

src/expressions/dataTypes/Value.ts
```diff
@@ -5,67 +5,390 @@ export default class Value {
...
+	MAP,
+	ARRAY,
+}
+
+/**
+ * Handles the occurances in the XPath specs.
+ * Zero or one matches to '?';
+ * One or more matches to '+';
+ * Zero or more matches to '*'
+ * @public
```
DrRataplan, line 84: nope, not public

DrRataplan (changes requested):

src/parsing/xpath.pegjs
```diff
@@ -1196,7 +1196,7 @@ SingleType
  = typeName:SimpleTypeName optional:"?"? {return optional ? ["singleType", ["atomicType"].concat(typeName), ["optional"]] : ["singleType", ["atomicType"].concat(typeName)]}
 
 // 183
-TypeDeclaration
+ValueType
```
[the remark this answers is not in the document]
HectorPeeters: We removed the original TypeDeclaration since it stored a type and a multiplicity which ValueType also did. ValueType is now renamed to SequenceType, so should we change it back to SequencType, or add TypeDeclaration back in?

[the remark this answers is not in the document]
DrRataplan: this is the parser, the TypeDeclaration string there is just something from the grammar of XPath/XQuery

DrRataplan:
Also, https://coveralls.io/builds/39525326/source?filename=src%2Fexpressions%2FdataTypes%2FValue.ts#L122 is marked as uncovered by the Code Coverage. This is usually an indication of dead code. Could you verify?

src/parsing/xpath.pegjs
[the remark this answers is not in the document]
HectorPeeters: Ooh my bad thought it was linked somehow

DrRataplan (approved):
Awesome job!

PR #349, opened 2021-05-21
Improve fuzzer

devatwork:
Improve the fuzzer:

1. Better crash reports;
2. Check for all known error codes;
3. Add XPath 3.1 function catalog.


DrRataplan:
## BundleMon

  
  
  
  
  
  
  
<details>
<summary>Unchanged files (2)</summary>

Status | Path | Size | Limits
:------------: | ------------ | :------------: | :------------:
:white_check_mark: | fontoxpath.esm.js<br/> | 88.72KB | -
:white_check_mark: | fontoxpath.js<br/> | 88.62KB | -
</details>



No change in files bundle size

Final result: :white_check_mark:

---
<p align="center"><a href="https://bundlemon.now.sh/projects/609024effb87a20009a6b096/reports?branch=improve-fuzzer&resolution=all" target="_blank" rel="noreferrer noopener">Current branch size history</a> | <a href="https://bundlemon.now.sh/projects/609024effb87a20009a6b096/reports?branch=master&resolution=days" target="_blank" rel="noreferrer noopener">Target branch size history</a></p>

fuzzers/fuzzer.ts
```diff
@@ -23,6 +54,20 @@ export class FuzzCase {
...
 
+	/**
+	 * Tests whether the given error is expected,
+	 * meaning the error should not be considered a crash.
+	 * 
+	 * @param error - The {@link Error} which to test.
+	 * @returns Returns `true` when expected.
+	 */
+	isExpectedError(error: Error): boolean {
+		return EXPECTED_ERROR_PREFIXES.some(prefix => error.message.startsWith(prefix));
```
jarivm, line 65: Maybe use a `Set`. It looks like all error messages have the same length, so something like `EXPECTED_ERROR_PREFIXED.has("ABCD000".slice(0, 4))` would work. Could give a small performance boost.

fuzzers/index.ts
```diff
@@ -29,10 +29,26 @@
...
-const engine = new Engine<CorpusBasedFuzzer>();
-engine.run(fuzzer, __filename);
+// Select the corpus to use
+const corpora: { [key: string]: ICorpusLoader } = {
+	[isoCorpusLoader.name]: isoCorpusLoader,
+	[xpath31FunctionCatalogCorpusLoader.name]: xpath31FunctionCatalogCorpusLoader
+};
+const corpusName = process.argv.slice(2)[0];
+const corpusLoader = corpora[corpusName];
+if (corpusLoader != null) {
```
jarivm, line 44: Missed a `!==` here.

fuzzers/xpath-31-function-catalog.xml
```diff
@@ -0,0 +1,18837 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jarivm, line 1: fontoxpath stores (big) test files in [test/assets/](https://github.com/FontoXML/fontoxpath/tree/master/test/assets). Please add instructions for downloading this file to the instructions for [running the fuzzer](https://github.com/FontoXML/fontoxpath/blob/master/CONTRIBUTING.md#running-the-fuzzer). This would make this file more maintainable. Maybe add a helpful error message when failing to load this document as well.

@DrRataplan do you agree with this?

fuzzers/fuzzer.ts
devatwork: Good point `O(1)` vs `O(n)` where `n` is the length of `EXPECTED_ERROR_PREFIXED`. Given the rest of the code, I doubt it will make an big impact on performance though. It should be measurable in a benchmark.

fuzzers/xpath-31-function-catalog.xml
devatwork: Why would you want to download that file? That would require writing code and thus raise maintenance cost. Also it is a point of failure and likely will lead to spurious build failures.

fuzzers/index.ts
devatwork: :barf: I don't know why my linter is still fubar, sad Js makes me sad. Will fix.

fuzzers/xpath-31-function-catalog.xml
DrRataplan: Where did you find this file? We indeed try to push these very large files to outside of this repo to keep git clones fast. If this is mirrored anywhere, I prefer to add instructions to fetch it from somewhere.

DrRataplan:
The fuzzer is getting large and complex. Some (automated) unit tests are in order, especially because this code is right now only executed by hand.

I'll give you a hand on that this Friday if you'd like?

fuzzers/xpath-31-function-catalog.xml
devatwork: https://www.w3.org/TR/2017/REC-xpath-functions-31-20170321/function-catalog.xml linked from https://www.w3.org/TR/xpath-functions/.

The file is ~135kb minified, so it is unlikely to significantly impact cloning speeds IMHO.

fuzzers/index.ts
DrRataplan: Needs to be fixed before integration

fuzzers/xpath-31-function-catalog.xml
DrRataplan: @MTimmerije could you add this to the asset download script (there are two now, in the `test` folder). That way we remove this file, it'll only get in the way of searching ad stuff. I do not expect us to make changes here. Don't forget to add it to the gitignore!

DrRataplan (changes requested):

MTimmerije:
> I'm not sure that adding unit tests make too much sense at this stage, the Fuzzer is still experimental IMHO, and unit tests will prevent us from iterating quickly.
> 
> Furthermore, breaking changes in the Fuzzer code, wont block any automated checks or releases or anyone from developing on Fontoxpath. So IMHO it is okay if its gets broken at some point in time.

I'll make some simple automated tests that just check that take queries from the corpus directly (without mutations). At least some simple tests like that would confirm that the basic functionality of the fuzzer works and that the queries from the corpus don't suddenly start to fail.

MTimmerije:
I had to actualize my branch using one of the more recent js codegen ones. 
So it contains a few more things than might be ideal.

DrRataplan:
> My changes brought some stuff from the jscodegen branch with it, as I integrated the BLNS to check if the jscodegen literals wouldn't misbehave. So it contains a few more things than just fuzzer changes.

In that case we'll first merge the js codegen changes and merge the fuzzer changes afterwards.

Issue #381, opened 2021-06-21
How to inject namespaced variables?


rrthomas:
As far as I can see from the source of `fontoxpath`, the `variables` parameter of `evaluateXPath` can only be used to bind variables in the global namespace. If I try binding a namespaced variable, it doesn't work:

```xquery
import {evaluateXPath, registerXQueryModule, Options} from 'fontoxpath'

const bar = 'https://example.com/bar.dtd'

registerXQueryModule(`module namespace bar = "${bar}";
declare variable $bar:var external;
declare function bar:baz() {
  $bar:var
};`)

const URI_BY_PREFIX: {[key: string]: string} = {bar}
const xQueryOptions: Options = {
  namespaceResolver: (prefix: string) => URI_BY_PREFIX[prefix],
  language: evaluateXPath.XQUERY_3_1_LANGUAGE,
  moduleImports: URI_BY_PREFIX,
}
const variables = {
  'bar:var': 'foo'
}

evaluateXPath('bar:baz()', null, null, variables, undefined, xQueryOptions)
```

Output:

```sh
$ npx ts-node ../src/bar.ts
XQDY0054: The variable var is declared but not in scope.
```

If I change `bar:var` to `var` everywhere, it works fine.

I cannot work out how to supply separately the value of a namespaced variable. I also tried two separate calls to `registerXQueryModule`, one containing the variable definition, and the other referring to it:

```typescript
import {evaluateXPath, registerXQueryModule, Options} from 'fontoxpath'

const foo = 'https://example.com/foo.dtd'
const bar = 'https://example.com/bar.dtd'

registerXQueryModule(`module namespace foo = "${foo}";
declare variable $foo:var := "hello world";`)

registerXQueryModule(`module namespace bar = "${bar}";
import module namespace foo = "${foo}";
declare function bar:baz() {
  $foo:var
};`)

const URI_BY_PREFIX: {[key: string]: string} = {foo, bar}
const xQueryOptions: Options = {
  namespaceResolver: (prefix: string) => URI_BY_PREFIX[prefix],
  language: evaluateXPath.XQUERY_3_1_LANGUAGE,
  moduleImports: URI_BY_PREFIX,
}

evaluateXPath(`bar:baz()`, null, null, null, undefined, xQueryOptions)
```

Output:

```sh
$ npx ts-node ../src/foo.ts
XPST0008, The variable var is not in scope.
```

[...]

PR #382, opened 2021-06-21
Improve documentation of ‘variables’ parameter to evaluateXPath

rrthomas:
Mention that only variables in the global namespace can be set.
(See issue #381.)


DrRataplan:
Awesome, thanks! Hope this will prevent some questions in the future.

PR #383, opened 2021-06-22
Implement namespaced variables in XQuery

### LINKED ISSUE #381: How to inject namespaced variables?

DrRataplan:
Fixes #381

### DIFF
--- a/test/specs/parsing/registerXQueryModule.tests.ts
+++ b/test/specs/parsing/registerXQueryModule.tests.ts
@@ -266,4 +266,57 @@ declare function x:fn () external;`);
 			)
 		).to.equal('meep');
 	});
+
+	it('can export variables with namespaces to a query', () => {
+		const foo = 'https://example.com/foo';
+
+		registerXQueryModule(`module namespace foo = "${foo}";
+declare variable $foo:var := "hello world";`);
+
+		const URI_BY_PREFIX: { [key: string]: string } = { foo };
+		const xQueryOptions = {
+			namespaceResolver: (prefix: string) => URI_BY_PREFIX[prefix],
+			language: evaluateXPath.XQUERY_3_1_LANGUAGE,
+			moduleImports: URI_BY_PREFIX,
+		};
+
+		chai.assert.equal(
+			evaluateXPath(
+				`import module namespace foo = "${foo}";
+$foo:var`,
+				null,
+				null,
+				null,
+				undefined,
+				xQueryOptions
+			),
+			'hello world'
+		);
+	});
+
+	it('can propagate variables with namespaces in modules', () => {
+		const foo = 'https://example.com/foo.dtd';
+		const bar = 'https://example.com/bar.dtd';
+
+		registerXQueryModule(`module namespace foo = "${foo}";
+declare variable $foo:var := "hello world";`);
+
+		registerXQueryModule(`module namespace bar = "${bar}";
+import module namespace foo = "${foo}";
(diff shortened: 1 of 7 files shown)


test/specs/parsing/registerXQueryModule.tests.ts
```diff
@@ -266,4 +266,30 @@ declare function x:fn () external;`);
 			)
 		).to.equal('meep');
 	});
+
+	it('can propagate variables with namespaces in modules', () => {
```
JosVerburg, line 270: I think we could also add another unit test which does the same but have it be a bit simpler:

```js
registerXQueryModule(`module namespace foo = "${foo}";
declare variable $foo:var := "hello world";`)

fontoxpath.evaluateXPath(`import module namespace foo = "${foo}";
	$foo:var`, null, null, null, undefined, xQueryOptions);

```

Should do the exact same but without providing a namespace resolver and importing in the query itself.

JosVerburg (approved):

PR #393, opened 2021-07-07
String literals jscodegen

HectorPeeters:
This pull request adds the initial support for string literal in the javascript codegen module. This will be used to implement compares in the future.

### DIFF
--- a/test/specs/jsCodegen/string.tests.ts
+++ b/test/specs/jsCodegen/string.tests.ts
@@ -15,7 +15,7 @@ describe('string tests', () => {
 
 	it('test simple string expression true', () => {
 		chai.assert.equal(
-			evaluateXPathWithJsCodegen("'hello'", documentNode, null, ReturnType.STRING, {}),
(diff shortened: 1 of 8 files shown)


test/specs/jsCodegen/string.tests.ts
```diff
@@ -0,0 +1,52 @@
+import * as chai from 'chai';
+import { ReturnType } from 'fontoxpath';
+import { sanitizeString } from 'fontoxpath/jsCodegen/escapeJavaScriptString';
+import * as slimdom from 'slimdom';
+import jsonMlMapper from 'test-helpers/jsonMlMapper';
+import evaluateXPathWithJsCodegen from './evaluateXPathWithJsCodegen';
+
+describe('string tests', () => {
```
DrRataplan, line 8: Can you add a test verifying evaluating a query like `child::element()` is not compiling to return a node by accident?

MTimmerije: Sure, I'll do that.

PR #395, opened 2021-07-13
Keep reference equality roundtripping maps and arrays through XPath

DrRataplan:
This can improve performance because a map does not have to be copied a second time when returning
from XPath. It also makes the APIs easier to use because an entry in the map will be exactly the
same as the entry that was put in to engine as a variable.

This makes FontoXPath more predictable.

Relates to discussion #392


DrRataplan:
This is a draft PR. I am aware naming is inconsistent. However, @JosVerburg @bwrrp, I think this PR can open up a lot of doors in fontoxpath-indices: more reference equality is usually more better.

bwrrp:
That's a lot of code changes, I'll try to take a look in more detail later.

Am I correct in assuming that the array / object will only retain reference equality as long as it has not been modified? Does that also apply to sub-values (e.g., a map stored in an array where the array is modified, but the map is not)? That could definitely be useful to prevent deep equality checks when using XQuery to manipulate such data!

DrRataplan:
Correct! It will keep equality of maps in check as long as they have not
been altered. As in, array:put accepts an array and a sequence and returns
a new array.

These changes will make sure that array is a new instance, but because of
the recursive nature of XDM to JavaScript object transformation, any
members *in* that array will be the same instances.

Same for maps.

Op di 13 jul. 2021 16:38 schreef Stef Busking ***@***.***>:

DrRataplan:
Seems like some part of the compiler is very cross with me indeed. Might be some of the `any` I removed.

PR #399, opened 2021-07-22
Turn the test asset installation commands into runnable scripts

rrthomas:


DrRataplan:
Awesome! Thanks a lot! Should make setting up the environment so much easier!

PR #411, opened 2021-08-31
[minor] Add a new ResultType.ALL_RESULTS that does no magic

DrRataplan:
The ANY result type is annoying. It returns a single node, empty arrays, filled arrays, objects,
etcetera.

This was a mistake from the beginning on. While it is very cool to have an XPath 'just work', it
causes bugs and the ugly pattern `if (!Array.isArray(result)) {result = [result])`. Also, attribute
nodes are atomized.

No longer! Just use the ALL_RESULTS type that does _what you expect_. This also deprecates the ANY
result type. FontoXPath 4.0 _will_ drop support for ANY!

### DIFF
--- a/test/specs/parsing/evaluateXPath.tests.ts
+++ b/test/specs/parsing/evaluateXPath.tests.ts
@@ -15,6 +15,7 @@ import {
 	evaluateXPathToString,
 	evaluateXPathToStrings,
 	getBucketForSelector,
+	ReturnType,
 } from 'fontoxpath';
 
 import jsonMlMapper from 'test-helpers/jsonMlMapper';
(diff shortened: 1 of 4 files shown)


test/specs/parsing/evaluateXPath.tests.ts
```diff
@@ -83,6 +93,90 @@ describe('evaluateXPath', () => {
...
+				evaluateXPath('.', documentNode, domFacade, {}, ReturnType.ALL_RESULTS),
+				[documentNode]
+			));
+		it('Keeps attributes attributes', () => {
+			const tmp = documentNode.createElement('tmp');
+			tmp.setAttribute('class', 'value');
+			chai.assert.deepEqual(
+				evaluateXPath('@class', tmp, domFacade, {}, ReturnType.ALL_RESULTS),
+				[tmp.getAttributeNode('class')]
+			);
```
bwrrp, line 134: Might be better to check that the nodes are actually the same object instead of deep equal (i.e., `chai.assert.sameMembers`)? If only because deepEqual will crash trying to traverse the cyclic structure if this somehow ends up a different instance.
(Applies to all the comparisons involving nodes)

bwrrp (approved):
LGTM, I left a few minor comments.

Good improvement to the API, especially once we can flip the default in 4.0. :)

PR #417, opened 2021-09-06
Implement fn:error() in one- and two-argument forms

rrthomas:


rrthomas:
Not quite sure what's going on with the tests: there are failures in the QT3 set in CI, but not when I run them on my machine.

With this addition (and `replace`), only `distinct-values` is needed to allow all but one of the functx functions to compile (this is not a particular goal I have, but it's a nice illustration of practical coverage).

DrRataplan:
Hey there,

It is very likely the [unrunnableTestCases.csv](https://github.com/FontoXML/fontoxpath/blob/master/test/assets/unrunnableTestCases.csv) file has outstanding changes on your machine. this file contains the list of tests that are known to fail. If it contains any new lines, those are tests that _have been known to work, but no longer do_. By removing those lines, or by running the tests using `npm run qt3tests -- --regenerate`, you can run them again. 

I have not read the test in depth yet, but it looks like our static analysis phase is recognizing that `empty(fn:error() is ())` is going to error out because the operands of the `is` are not nodes: they are the empty sequence (which is OK), and the return of the `error` function, which is never going to happen. I am unsure what the spec actually thinks of this.

Hope this helps.

rrthomas:
Thanks, @DrRataplan, that was indeed the case; I'll look into it again.

rrthomas:
I checked out the `unrunnableTestCases.csv` file so that it did not contain any local modifications and reran `npm run qt3tests`, which ran fine.

I then ran `npm run qt3tests -- --regenerate`, which indeed modified `unrunnableTestCases.csv`. I studied the results, which seem mostly positive.

The apparent regressions are:

```
+fn-error-6,AssertionError: expected [Function] to throw error matching /FOER0000/ but got 'Not implemented: Using an error obje
ct in error is not supported'
+cbcl-error-021,AssertionError: Expected executing the XPath "typeswitch ( fn:error() ) case xs:integer return true() default re
turn false()" to resolve to one of the expected results, but got Error: XPST0003: Unable to parse: "typeswitch ( fn:error() ) ca
se xs:integer return true() default return false()". Expected " ", "(:", "\n", "\r", "\t", or "function" but "f" found. typeswit
ch ( [Error is around here]fn:error() ) case xs:integer return true() default return false(), AssertionError: expected [Function
] to throw error matching /FOER0000/ but got 'XPST0003: Unable to parse: "typeswitch ( fn:error() ) case xs:integer return true(
) default return false()".\nExpected " ", "(:", "\\n", "\\r", "\\t", or "function" but "f" found.\ntypeswitch ( [Error is around
 here]fn:error() ) case xs:integer return true() default return false()'.
+cbcl-error-022,AssertionError: Expected executing the XPath "typeswitch ( fn:error() ) case xs:integer return true() default re
turn false()" to resolve to one of the expected results, but got Error: XPST0003: Unable to parse: "typeswitch ( fn:error() ) ca
se xs:integer return true() default return false()". Expected " ", "(:", "\n", "\r", "\t", or "function" but "f" found. typeswit
ch ( [Error is around here]fn:error() ) case xs:integer return true() default return false(), AssertionError: expected [Function
] to throw error matching /FOER0000/ but got 'XPST0003: Unable to parse: "typeswitch ( fn:error() ) case xs:integer return true(
) default return false()".\nExpected " ", "(:", "\\n", "\\r", "\\t", or "function" but "f" found.\ntypeswitch ( [Error is around
 here]fn:error() ) case xs:integer return true() default return false()'.
+cbcl-error-028,AssertionError: Expected executing the XPath "declare function local:item() { if (current-date() lt xs:date('201
2-10-10')) then 1 else "one" }; empty( typeswitch ( local:item() ) case xs:integer return fn:error() default return ())" to reso
lve to one of the expected results, but got Error: XPST0003: Unable to parse: "declare function local:item() { if (current-date(
) lt xs:date('2012-10-10')) then 1 else "one" }; empty( typeswitch ( local:item() ) case xs:integer return fn:error() default re
turn ())". Expected " ", "(:", "\n", "\r", "\t", or "function" but "l" found. declare function local:item() { if (current-date()
 lt xs:date('2012-10-10')) then 1 else "one" }; empty( typeswitch ( [Error is around here]local:item() ) case xs:integer return 
fn:error() default return ()), AssertionError: expected [Function] to throw error matching /FOER0000/ but got 'XPST0003: Unable 
to parse: "declare function local:item() { if (current-date() lt xs:date(\'2012-10-10\')) then 1 else "one" }; empty( typeswitch
 ( local:item() ) case xs:integer return fn:error() default return ())".\nExpected " ", "(:", "\\n", "\\r", "\\t", or "function"
 but "l" found.\ndeclare function local:item() { if (current-date() lt xs:date(\'2012-10-10\')) then 1 else "one" }; empty( type
switch ( [Error is around here]local:item() ) case xs:integer return fn:error() default return ())'.
+cbcl-error-029,Error: XPST0003: Unable to parse: "declare function local:item() { if (current-date() gt xs:date('1900-01-01')) 
then 1 else "one" }; empty( typeswitch ( local:item() ) case xs:integer return () default return fn:error())". Expected " ", "(:
", "\n", "\r", "\t", or "function" but "l" found. declare function local:item() { if (current-date() gt xs:date('1900-01-01')) t
hen 1 else "one" }; empty( typeswitch ( [Error is around here]local:item() ) case xs:integer return () default return fn:error()
)
+cbcl-error-035,AssertionError: Expected executing the XPath "empty( fn:error() treat as empty-sequence() )" to resolve to one of the expected results, but got Error: No selector counterpart for: treatExpr., AssertionError: expected [Function] to throw error matching /FOER0000/ but got 'No selector counterpart for: treatExpr.'.
+xs-error-047,AssertionError: expected [Function] to throw error matching /FOER0000/ but got 'No selector counterpart for: treatExpr.'
+K-SeqExprInstanceOf-52,AssertionError: expected [Function] to throw error matching /XPST0051/ but got 'Cannot read property \'v
ariety\' of undefined'
+K-SeqExprInstanceOf-58,AssertionError: Expected executing the XPath "error() instance of empty-sequence()" to resolve to one of
 the expected results, but got Error: No selector counterpart for: voidSequenceType., AssertionError: expected [Function] to thr
ow error matching /FOER0000/ but got 'No selector counterpart for: voidSequenceType.'.
+K-SeqExprInstanceOf-59,AssertionError: Expected executing the XPath "(error(), 1) instance of empty-sequence()" to resolve to o
ne of the expected results, but got Error: No selector counterpart for: voidSequenceType., AssertionError: expected [Function] t
o throw error matching /FOER0000/ but got 'No selector counterpart for: voidSequenceType.'.
```

rrthomas:
I tried to fix `K-SeqExprInstanceOf-52`: it seemed that it was caused by the lack of an entry in `builtinModels`, so I added one:

```
	{
		variety: Variety.UNION,
		name: ValueType.NONE,
		memberTypes: [],
	},
```

since according to https://www.w3.org/TR/xquery-semantics/ "none is not a FormalItemType, rather it is equivalent to a union type with no member types". However, this did not fix the failing test.

DrRataplan:
I found it: the NONE type is not really there. It is not usable by users, so it should not be in the `stringToValueTypeMap`. However, it should also not be used for type inference.

I pushed a commit that should make this a lot more stable

DrRataplan:
The coveralls tooling was being quite unstable the other day. Kicked all the pipelines awake again, this should be shippable when they finish. Afterwards I'll ship a new fontoxpath

PR #421, opened 2021-09-18
Refer to 'function' not 'test' in documentation for registerCustomXPathFunction

rrthomas:


DrRataplan:
Awesome, thanks! This was very old: FontoXPath used to be a CSS-like selector engine. You can even see that in the first commits! We slowly grew to the XPath/XQuery engine we have now

PR #422, opened 2021-09-19
Implement one-argument form of fn:serialize()

rrthomas:


rrthomas:
Note: I do not even pretend this works yet!

I got stuck trying to make closure happy: as with xspatterns, I made slimdom a run-time dependency, but there seems to be a problem with type annotations; I get 4 warnings like the following:

```
ClosureCompiler: /home/rrt/.local/var/repo/fontoxpath/.tscc_temp/50a970eb1f9d4179737a70bb05d4ead23d4e1021b022b47f2580dce735744e9f/externs_generated.js:657:67: WARNING - [JSC_TYPE_PARSE_ERROR] Bad type annotation. expected closing } See https://github.com/google/closure-compiler/wiki/Annotating-JavaScript-for-the-Closure-Compiler for more information.
  657|  * @implements {node_modules$slimdom$dist$slimdom$$d$$ts.ParentNode|node_modules$slimdom$dist$slimdom$$d$$ts.NonDocumentTypeChildNode|node_modules$slimdom$dist$slimdom$$d$$ts.ChildNode}
```

This seems to relate to: https://ask.clojure.org/index.php/6240/jsdoc-parsing-fails-recognize-union-types-breaking-resolution

As a result (I guess), when I try a simple test I get:

```
slimdom is not defined
```

Of course, I don't know whether my actual implementation is any good either! I hope that it's fairly simple, anyway.

DrRataplan:
It seemed like such a great idea: an XPath engine without any hard dependency on any specific DOM implementation; as long as it implements the DOM interfaces, we accept it. This has not been making development easy, but I still do not want to give it up. Having an XPath implementation that can work with the native DOM (in the browser) as well as lightweight implementations (like slimdom) is very powerful.

This means we should either implement serialization in FontoXPath, or find a dependency that does the minimal part of it, or make it injectable: an extra option XMLSerializer that can be passed when calling `evaluateXPath`. The XMLSerializer packed with slimdom does not like nodes that it did not create itself, so that is a dead path. 

I think I feel most for an option `xmlSerializer`, that will be placed on the `evaluationParameters`, next to the `documentWriter` and `nodesFactory`.

rrthomas:
Thanks for the design guidance, @DrRataplan, I'll have a go at implementing that. Also, thanks for the hint re slimdom; I guess I'll have to find another serializer for my application, but it seems there are some to choose from.

rrthomas:
I've implemented this as you suggest, @DrRataplan, and a basic test with `xmlserializer` works (I note that `w3c-xmlserializer` might be a good option, but it doesn't implement `XMLSerializer` out of the box, so would need a wrapper.)

I have pushed that; I'll now see if I can wire up the fontoxpaths to also use `xmlserializer`.

rrthomas:
Some of the tests now seem to be passing; unfortunately not most, as they require the two-argument form of `serialize()`. I haven't looked, but it may be that some of those would pass with a trivial implementation!

rrthomas:
@DrRataplan Thanks very much for your help with this!

DrRataplan:
Hey there,

I just pushed some changes on top of your PR: I declared some better local typing, fixed some linting issues and I added some unit tests. Lets see how the CI likes the new coverage for tests. The error cases are often missed in the normal QT3 tests.

TODO: verify coverage, finish the public API part and consider to implement (a part of) the two-argument version.

Thanks again for this! This collaboration is very helpful!

Regards,

Martin

rrthomas:
I just tried using this code with your changes, @DrRataplan, and I found a couple of things I don't understand:

1. When setting the `xmlSerializer` property, I have to write something like `xmlSerializer: new slimdom.XMLSerializer() as fontoxpath.XMLSerializer`. The tests use slimdom.XMLSerializer without this cast, so I'm not sure why I need it.
2. `serialize()` seems to return something that isn't a string. I can't work out what's going wrong from looking at the source code, but when I pass the result of `serialize()` to my own user function that declares its argument type to be `xs:string`, with a corresponding TypeScript signature of `string`, the actual argument is an `object`, and if I call `.toString()` I get the value I expect.

Sorry I don't have minimal failing examples yet, I'm just trying to get some code working, and working around as above is good enough for me for now. Feel free to ignore this report until I can supply a proper example!

rrthomas:
I have refreshed this PR on top of current master. I am delighted to see that `distinct-values` is now implemented, which means that all of functx now compiles except for `functx:open-ref-document()`, which needs `base-uri()`.

DrRataplan (approved):

DrRataplan:
Hey Reuben,

This seems to have gotten a bit snowed over I'm afraid! The implementation looks AOK to me and I'm very happy we seem to be close to 'supporting' the functx library! To be honest I would've never thought we would ever get close!

That base-uri function does not seem too hard to implement at first sight, for the browser DOM that is. However I see in Ruth you are using slimdom, which does not support the baseURI property on Node. I really wonder how that is supposed to work though in for example NodeJS where your files might not even come from disk. Lets see, I'm always happy to implement `fn:base-uri` to returning `about:blank` or something for any DOM implementation that does not support base-uri, but that is very silly.

I'm integrating this now and I'm planning to make a new release this week. Thanks for the help!

Kind regards,

Martin

rrthomas:
Fantastic that this is merged, I will no longer need a fork of fontoxpath for Ruth!

I don't need to use `base-uri()`, but it would be great if FunctX compiled without needing to comment out its use of that function, so a stub implementation is fine by me.

PR #437, opened 2021-11-04
Implementation of distinct-values

fhaftmann:
* The implementation keeps the first occurrence in a sequence – this is not demanded by the specification, but some tests are written that way.
* The implementation uses a naive nested loop; a more optimized version would require a concept of hash values or similar to partition elements into buckets – I did not find something like that in the implementation yet and decided not to dive into such a fundamental issue of runtime presentation.
* There seems to be a weakness in comparing durations, which I will address in a separate pull request.


DrRataplan:
Hey there! Thanks you so much for implementing this function! I agree that the nested loop may be a bit naive, but I think that for now, this implementation should suffice. We can always address any performance issues later: first make it work, then make it fast.

I'll take some time fixing the linter and anything else the CI might find.

Again, thanks so much!

DrRataplan:
Just rebased it and I updated some of the unrunnableTests CSV files. I will merge this this week so it can be shipped with the version of fontoXPath I'm planning to ship at the end of the week.

DrRataplan (approved):

PR #438, opened 2021-11-04
More correct comparision of durations

fhaftmann:
This pull request continues after https://github.com/FontoXML/fontoxpath/pull/437.

It adds a spec-conform comparison for durations, which seemed to be absent before.

The issue came to surface when implementing distinct-values.


DrRataplan:
This got integrated together with the #437 PR. Many many thanks for this!

PR #448, opened 2021-11-29
DEV-12516 Use strongest bucket for AndExpression rather than the first one

bwrrp:
This improves the `AndExpression` `getBucket` method to compare buckets from the subexpressions and generate a bucket that represents their intersection. As there isn't currently a bucket to represent a selector that never matches, this introduces `'empty'` to represent that case.

### DIFF
--- a/src/expressions/operators/boolean/AndOperator.ts
+++ b/src/expressions/operators/boolean/AndOperator.ts
@@ -11,6 +11,44 @@ import Expression from '../../Expression';
 import Specificity from '../../Specificity';
 import { DONE_TOKEN, ready } from '../../util/iterators';
 
+function intersectBuckets(bucket1: string | null, bucket2: string | null): string | null {
+	// null bucket applies to everything
+	if (bucket1 === null) {
+		return bucket2;
+	}
+	if (bucket2 === null) {
+		return bucket1;
+	}
+	// empty bucket applies to nothing
+	if (bucket1 === 'empty' || bucket2 === 'empty') {
+		return 'empty';
+	}
+	// Same bucket is same
+	if (bucket1 === bucket2) {
+		return bucket1;
+	}
+
+	if (bucket1.startsWith('name-')) {
+		// Name bucket always refers to an element or attribute
+		if (bucket2 === 'type-1' || bucket2 === 'type-1-or-type-2' || bucket2 === 'type-2') {
+			// A name is more specific than a type
+			return bucket1;
+		}
+		// Even if bucket2 is a name, we know it's not equal, so intersection is empty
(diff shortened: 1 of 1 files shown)


DrRataplan:
Good idea! And thanks for the implementation! I am missing some tests for this though. We already have some unit tests in place for buckets in the `and` case, they should be easy to extend.

src/expressions/operators/boolean/AndOperator.ts
```diff
@@ -11,6 +11,44 @@ import Expression from '../../Expression';
...
+	}
+
+	if (bucket1.startsWith('name-')) {
+		// Name bucket always refers to an element or attribute
+		if (bucket2 === 'type-1' || bucket2 === 'type-1-or-type-2' || bucket2 === 'type-2') {
+			// A name is more specific than a type
+			return bucket1;
+		}
+		// Even if bucket2 is a name, we know it's not equal, so intersection is empty
+		return 'empty';
```
DrRataplan, line 38: This bucket is new. Won't we have any code in Fonto that will break because of it? In theory, other users can also use buckets (though I heavily doubt they do).

Also, when the bucket really _is_ `'empty'`, we can compile the `and` expression to a no-op that returns `false`, but that's for later.

bwrrp:
> I am missing some tests for this though.

I've added a few assertions in `getBucketForSelector.tests.ts` to hopefully cover all branches.

src/expressions/operators/boolean/AndOperator.ts
bwrrp: IIRC, previous problems were caused by new buckets having overlap with existing ones. Empty by definition doesn't overlap anything, so I'm not expecting problems. It will probably cause selectors that have this bucket to never be evaluated for retrieving metadata, which seems fine if they never match anyway.
In addition to a no-op, it might be worth warning in dev mode if this happens - any selector that doesn't match any nodes (aside from `false()`) is probably unintentional.

DrRataplan (approved):

src/expressions/operators/boolean/AndOperator.ts
```diff
@@ -11,8 +11,45 @@ import Expression from '../../Expression';
 import Specificity from '../../Specificity';
 import { DONE_TOKEN, ready } from '../../util/iterators';
 
+const subBucketsByBucket: Record<string, string[]> = {
```
JosVerburg, line 14: Maybe comment what this expresses. I don't think the name `subBucketsByBucket` indicates very clearly that it gives you buckets which are more specific

JosVerburg (approved):

PR #453, opened 2021-12-01
Make fontoxpath:evaluate accept an AST as well as a string

DrRataplan:
This is required for a special case in Fonto: we sometimes retrieve a query from some registry before passing it to `fontoxpath:evaluate`. Now that we are working more and more with preparsed queries, `fontoxpath:evaluate` needs to be able to handle them.

### DIFF
--- a/src/expressions/functions/builtInFunctions_fontoxpath.ts
+++ b/src/expressions/functions/builtInFunctions_fontoxpath.ts
@@ -1,28 +1,28 @@
-import convertXmlToAst from '../../parsing/convertXmlToAst';
+import { ElementNodePointer } from 'src/domClone/Pointer';
 import { printAndRethrowError } from '../../evaluationUtils/printAndRethrowError';
 import astHelper, { IAST } from '../../parsing/astHelper';
 import compileAstToExpression from '../../parsing/compileAstToExpression';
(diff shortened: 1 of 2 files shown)


src/expressions/functions/builtInFunctions_fontoxpath.ts
```diff
@@ -1,23 +1,140 @@
...
+			allowXQuery: false,
+			debug,
+		});
+	}
+
+	if (isSubtypeOf(queryValue.type, ValueType.ELEMENT)) {
+		const nodePointer: ElementNodePointer = queryValue.value;
+
+		try {
+			return convertXmlToAst(nodePointer.node);
```
bwrrp, line 39: Should this be able to handle a mix of nodes constructed in the query itself and (clones of) nodes existing in the context? That would enable modifying XQuery using XQuery, which is kind of neat. It seems that by taking the node out of the pointer we're discarding such information. I'm fine with leaving that out for now, but I do think in that case we should throw if the nodePointer is a structure that has any grafts.

bwrrp (approved):

src/expressions/functions/builtInFunctions_fontoxpath.ts
DrRataplan: True, but lets worry about that when we get there.

PR #457, opened 2022-01-14
Prsc parser

HectorPeeters:
This pull request replaced the pegjs parser with a parser written using the prsc framework.


src/parsing/prscParser.ts
```diff
@@ -0,0 +1,2795 @@
...
+
+	return surrounded(module, whitespace);
+}
+
+export function parseUsingPrsc(
+	xpath: string,
+	options: { outputDebugInfo: boolean; xquery: boolean },
+	shouldConsumeAll: boolean = true
+): ParseResult<IAST> {
+	return (shouldConsumeAll ? complete(generateParser(options)) : generateParser(options))(
```
DrRataplan, line 2791: Besides the TODOs in this file, lets try to not call this `generateParser` runTime. there are only going to be two or three unique versions anyway: the XPath one, the XQuery one and an XQuery Update Facility one.

also, the shouldConsumeAll is I think not really something we'll ever set to `false`, right?

src/parsing/prscParser.ts
```diff
@@ -0,0 +1,2795 @@
...
+
+	const mainModule: Parser<IAST> = then(
+		prolog,
+		preceded(whitespace, queryBody),
+		(prologPart, body) => ['mainModule', ...(prologPart ? [prologPart] : []), body]
+	);
+
+	const versionDeclaration: Parser<IAST> = map(
+		precededMultiple(
+			[token('xquery'), whitespace],
```
DrRataplan, line 2747: It might be cool to deduplicate these tokens: add something (maybe even another file to make this one a bit smaller?) that defines them: `const XQUERY_TOKEN = token('xquery'); const BRACE_OPEN_TOKEN = token('(');` etcetera?

src/parsing/prscParser.ts
```diff
@@ -0,0 +1,2795 @@
...
+			)
+		),
+		(x) => ['versionDecl', ...x]
+	);
+
+	const module: Parser<IAST> = then(
+		optional(surrounded(versionDeclaration, whitespace)),
+		or([libraryModule, mainModule]),
+		(versionDecl, modulePart) =>
+			['module', ...(versionDecl ? [versionDecl] : []), ...[modulePart]] as IAST
```
DrRataplan, line 2780: Cool to see all these tokens in TypeScript now. Maybe in the future we can something akin to what y'all did earlier: replace the strings with number constants / enums. While parsing is not a significant part of the performance of the engine, it might remove the risk of some silly typos in silly places.

Not for now though!

src/parsing/prscParser.ts
HectorPeeters: The `shoudConsumeAll` was an option used during debugging which can indeed be removed. As for calling `generateParser` at runtime, I have not yet found a way to easily clear all the caches of the specific parsing functions.

DrRataplan (approved):

PR #469, opened 2022-02-02
Annotate namespace URIs when they are present in the prolog

DrRataplan:
This will make folding XQuery modules into eachother possible


DrRataplan:
The changes in the `failingXQueryXTestNames` are expected: they are parsed, but unsupported. Better to throw an error early

PR #481, opened 2022-03-04
Cache expressions in fontoxpath:evaluate

bwrrp:
This removes a bit of duplicated code and instead makes `fontoxpath:evaluate` use `staticallyCompileXPath`. That enables caching for expressions evaluated this way and avoids unnecessary parsing.

### DIFF
--- a/src/expressions/functions/builtInFunctions_fontoxpath.ts
+++ b/src/expressions/functions/builtInFunctions_fontoxpath.ts
@@ -60,8 +60,7 @@ function buildResultIterator(
 				allowUpdating: false,
 				allowXQuery: true,
 				debug: executionParameters.debug,
-				// TODO: should we inherit this from the outer evaluation somehow?
-				disableCache: false,
+				disableCache: executionParameters.disableCache,
 			},
 			(prefix) => staticContext.resolveNamespace(prefix),
 			// Set up temporary bindings for the given variables
--- a/src/expressions/functions/builtInFunctions_fontoxpath.ts
+++ b/src/expressions/functions/builtInFunctions_fontoxpath.ts
@@ -1,12 +1,6 @@
-import { ElementNodePointer } from '../../domClone/Pointer';
+import { EvaluableExpression } from '../..';
 import { printAndRethrowError } from '../../evaluationUtils/printAndRethrowError';
-import astHelper, { IAST } from '../../parsing/astHelper';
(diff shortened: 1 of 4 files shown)


src/parsing/convertXmlToAst.ts
```diff
@@ -14,7 +15,7 @@ export default function convertXmlToAst(element: Element): IAST {
 		element.namespaceURI !== BUILT_IN_NAMESPACE_URIS.FONTOXPATH_NAMESPACE_URI &&
 		element.namespaceURI !== BUILT_IN_NAMESPACE_URIS.XQUERYX_UPDATING_NAMESPACE_URI
 	) {
-		throw new Error('AST should only contain XQUERYX elements');
+		throw errXPTY0004('The XML structure passed as an XQueryX program was not valid XQueryX');
```
bwrrp, line 18: Not entirely sure about this one, the fontoxpath:evaluate test expected XPTY0004, and it seems that there were no other tests expecting anything else if an incorrect XQueryX AST is passed directly to evaluateXPath...

src/expressions/functions/builtInFunctions_fontoxpath.ts
```diff
@@ -69,64 +53,41 @@ function buildResultIterator(
...
-
-	const innerDynamicContext = new DynamicContext(context);
+		const { expression, staticContext: innerStaticContext } = staticallyCompileXPath(
+			getEvaluableExpressionFromValue(queryValue),
+			{
+				allowUpdating: false,
+				allowXQuery: true,
+				debug: executionParameters.debug,
+				// TODO: should we inherit this from the outer evaluation somehow?
+				disableCache: false,
```
bwrrp, line 64: It might be good to disable this cache if the cache was disabled for the outer evaluation, but it doesn't seem like information this is currently passed along anywhere. Should I add this property to executionParameters?

src/parsing/convertXmlToAst.ts
DrRataplan: Sounds like a proper error code: some type is unexpected

src/expressions/functions/builtInFunctions_fontoxpath.ts
DrRataplan: Cool, sounds great

DrRataplan (approved):

PR #482, opened 2022-03-24
Make the child:: axis lazy

DrRataplan:
This helps in performance an in getting the minimal dependencies for queries that only want to see
the first child


test/specs/parsing/axes/ParentAxis.tests.ts
```diff
@@ -87,8 +87,8 @@ describe('parent', () => {
 				getParentNode(node: Node, _bucket?: string | null): Node | null {
 					return node.parentNode;
 				},
-				getChildNodes(node: Node, _bucket?: string | null): Node[] {
-					return node.childNodes;
+				getFirstChild: (node, bucket: string | null) => {
+					return null;
 				},
```
bwrrp, line 92: nit: might be better to write this as a method instead of an arrow function for consistency?

bwrrp (approved):

PR #515, opened 2022-07-28
Improve trace message

JosVerburg:
Do note that I've changed the following to our trace logging:
* Nodes are now serialized if a serialized is provided in the options
* We now log the type (and now type string) of the original value instead of the atomized value which was always cast to a string.

### DIFF
--- a/src/expressions/functions/builtInFunctions_debugging.ts
+++ b/src/expressions/functions/builtInFunctions_debugging.ts
@@ -1,7 +1,10 @@
+import { NodePointer } from '../../domClone/Pointer';
+import realizeDom from '../../domClone/realizeDom';
 import atomize from '../dataTypes/atomize';
 import castToType from '../dataTypes/castToType';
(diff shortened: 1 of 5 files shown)


src/expressions/functions/builtInFunctions_debugging.ts
```diff
@@ -14,18 +17,24 @@ const fnTrace: FunctionDefinitionType = (
...
+							.map((atomizedValue) => castToType(atomizedValue, ValueType.XSSTRING))
+							.first().value;
+
 			newMessage +=
-				'{type: ' +
-				argumentAsStrings[i].type +
-				', value: ' +
-				argumentAsStrings[i].value +
-				'}\n';
+				'{type: ' + valueTypeToString(value.type) + ', value: ' + argumentAsString + '}\n';
```
DrRataplan, line 37: Can you rewrite this to a backtick string?

DrRataplan (approved):

Issue #509, opened 2022-07-11
Improve error message when evaluating a query AST


JosVerburg:
When we evaluate a query AST we don't get as good error messaging as with normal string queries.

I've reproduced this in the `./demo/main.ts` by changing the `runNormalXPath` to

```ts
async function runNormalXPath(script: string, asXQuery: boolean) {
	const queryAst = fontoxpath.parseScript(
		script,
		{
			language: fontoxpath.Language.XQUERY_3_1_LANGUAGE,
			debug: false,
		},
		new Document()
	);

	const raw = [];
	const it = fontoxpath.evaluateXPathToAsyncIterator(queryAst, xmlDoc, null, null, {
		debug: true,
		disableCache: true,
		language: asXQuery
			? fontoxpath.evaluateXPath.XQUERY_3_1_LANGUAGE
			: fontoxpath.evaluateXPath.XPATH_3_1_LANGUAGE,
		logger: {
			trace: (m) => {
				traceOutput.textContent = m;
				console.log(m);
			},
		},
	});
	...
```

I know that I do set the `debug` flag to `false` when building the AST, but the evaluate function throws an `invalid array length` error when I set it to `true` instead.

Using this, when I execute the following XQuery `if ((1, 2)) then '（＾∀＾●）ﾉｼ' else '╰（‵□′）╯'` I get the following error message

```
FORG0006: A wrong argument type was specified in a function call.
```

While I was expecting the error message as I would get it when treating the query as a string (pass the `script` to `evaluateXPathToAsyncIterator` instead of `queryAst`)

```
1: if ((1, 2)) then '（＾∀＾●）ﾉｼ' else '╰（‵□′）╯'
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
2: 

Error: FORG0006: A wrong argument type was specified in a function call.
  at <ifThenElseExpr>:1:1 - 1:43
```

PR #519, opened 2022-08-03
Better trace

DrRataplan:


src/parseScript.ts
```diff
@@ -129,7 +129,12 @@ function parseNode(
 					}
 					continue;
 				}
-
+				if (
+					attributeName === 'start' ||
```
JosVerburg, line 133: Should the parenthesis not be around both attribute names?
Like:
```
(attr.localName === 'start' || attr.localName === 'end')
&& element.localName === 'stackTrace'
```

JosVerburg (approved):
Thanks, works like a charm to fix #509

PR #549, opened 2022-10-13
Codegen improvements

bwrrp:
This started out as implementing codegen for the not() function, but grew into a little bit more...

* fixed fn:name function not returning empty string for empty sequence argument
* refactored codegen to generate mostly expressions and automatically manage identifiers rather than having to assign identifiers to everything - this reduced the generated code types to value, generator (paths only) or statement (generator body)
* refactored codegen to pass and combine PartialCompilationResults everywhere instead of plain strings
* aligned conversion functions to spec terminology (effective boolean value, atomization) and checked for correctness
* fixed several instances of variables not being passed / combined (made variables required to reduce errors)
* fixed a performance issue where the child axis would call getChildNodes for each iteration of the generator - it now loops using getFirstChild and getNextSibling
* fixed some cases where buckets were not passed correctly to axis domFacade calls
* fixed some issues in type annotation
* implemented general comparison for sequence vs single value in order to not regress for attribute = value comparisons (previously only accepted because multiplicity for the attribute axis was not annotated correctly)
* implemented codegen support for the context item expression
* implemented codegen support for the node() test
* implemented codegen support for non-axis path steps (filterExpr)
* implemented codegen support for calling custom functions
* implemented codegen support for calling selected built-in functions
* implemented native codegen for the not() function

Things to note:
* Built-in functions use an allowlist, as codegen'ed function calls don't currently have the full staticContext, executionParameters etc.
* I had to add an options argument to executeJavascriptCompiledXPath in order to pass currentContext (required by custom functions)
* I disabled the no-shadowed-variables tslint rule, because I think renaming the callback argument in the many mapPartialCompilationResult to something other than the var being mapped makes the code confusing and may even invite errors where the callback body tries to use the unchecked expression instead of the argument.


src/jsCodegen/runtimeLib.ts
```diff
@@ -1,66 +1,77 @@
...
-	isSubtypeOf: typeof isSubtypeOf;
-	ready: typeof ready;
+	callFunction: (
+		domFacade: IDomFacade,
+		namespaceURI: string,
+		localName: string,
+		args: UntypedExternalValue[],
+		options: Options | null
+	) => unknown;
+	XPDY0002: typeof XPDY0002;
```
DrRataplan, line 23: Where is this used?

src/jsCodegen/emitFunctionCallExpr.ts
```diff
@@ -13,121 +30,286 @@ const supportedFunctions: Record<
...
+	contextItemExpr: PartialCompilationResult,
+	context: CodeGenContext
+): PartialCompilationResult {
+	return mapPartialCompilationResult(
+		context.getIdentifierFor(contextItemExpr, 'contextItem'),
+		(contextItemExpr) =>
+			acceptAst(contextItemExpr.code, { type: GeneratedCodeBaseType.Value }, [
+				...contextItemExpr.variables,
+				`if (${contextItemExpr.code} === undefined || ${contextItemExpr.code} === null) {
+					throw new Error('XPDY0002: The function which was called depends on dynamic context, which is absent.');
```
DrRataplan, line 235: Why are you not using the runtimelib here?

DrRataplan (changes requested):
Awesome work! Especially like the yield statements! Makes the code way easier to read.

Left a few small nitpicky comments

src/jsCodegen/emitAxis.ts
```diff
@@ -0,0 +1,230 @@
...
+			}) : [])`,
+			{ type: GeneratedCodeBaseType.Value },
+			contextItemExpr.variables
+		)
+	);
+	const iterationCode = mapPartialCompilationResult(stepContextItemExpr, (stepContextItemExpr) =>
+		mapPartialCompilationResult(nodesExpr, (nodesExpr) =>
+			acceptAst(
+				`const ${stepContextItemExpr.code} of ${nodesExpr.code}`,
+				{ type: GeneratedCodeBaseType.Statement },
```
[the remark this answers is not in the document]
bwrrp: Yep, or rename Statement to NonExpressionCodeFragment or something... It's only used here so felt like overkill to introduce a separate type...

src/jsCodegen/emitFunctionCallExpr.ts
bwrrp: To this and the above - error throwing was and still is a bit inconsistent - do you prefer to import all error helpers through runtimeLib or to turn everything into `new Error('XP...')`?

DrRataplan (approved):
Awesome! Let me know what the performance results do for the queries that are now supported by codegen!

PR #557, opened 2022-11-08
Allow adding a seed for fn:random-number-generator

DrRataplan:


src/expressions/DynamicContext.ts
```diff
@@ -83,6 +86,20 @@ class DynamicContext {
 		return this._temporalContext.implicitTimezone;
 	}
 
+	/**
+	 * Returns a pseudo-random integer and a decimal based on the passed seed
+	 */
+	public getRandomNumber(seed: number = this._defaultSeed): {currentInt: number, currentDecimal: number} {
+		// Seed a prng with the chosen seed. This is a linear congruential generator
+		const m = 134456;
```
bwrrp, line 94: Is having 134456 different floats enough?

Additionally, does the spec define the range of random ints anywhere?

src/expressions/functions/builtInFunctions_numeric.ts
```diff
@@ -303,86 +303,113 @@ const fnNumber: FunctionDefinitionType = (
...
-	// Use Durstenfeld shuffle to randomize the list
-	for (let i = a.length - 1; i > 1; i--) {
-		const j = Math.floor(Math.random() * (i + 1));
-		const t = a[j];
-		a[j] = a[i];
-		a[i] = t;
+function createSeedFromString (seed: string): number {
+	let iseed = 0;
+	for (let i = 0; i < seed.length; ++i) {
+		iseed += seed.charCodeAt(i);
```
bwrrp, line 309: Probably ok, but this "hash" might be a bit primitive - order of the characters in the string does not affect the hash and this always favors lower integers for smaller strings. Probably lots of other collisions there. No idea how that affects the randomness...

src/expressions/functions/builtInFunctions_numeric.ts
```diff
@@ -303,86 +303,113 @@ const fnNumber: FunctionDefinitionType = (
...
-				value: () =>
-					sequenceFactory.singleton(
-						new FunctionValue({
-							value: permute,
-							isAnonymous: true,
-							localName: '',
-							namespaceURI: '',
-							argumentTypes: [
-								{
+	debugger
```
bwrrp, line 320: guessing this should have been removed?

src/expressions/DynamicContext.ts
DrRataplan: I added a way better PRNG algorithm now, with a larder range.

src/expressions/functions/builtInFunctions_numeric.ts
DrRataplan: I added a better hashing algorithm here

src/expressions/functions/builtInFunctions_numeric.ts
DrRataplan: Done

PR #561, opened 2022-11-30
Circ deps

DrRataplan:
This adds modules that can be spread out over multiple files. With function references going back and forth!


src/parsing/processProlog.ts
```diff
@@ -217,16 +233,14 @@ function processFunctionDefinition(
...
-			// Only mark the registration as the public API for the module if it's public
-			compiledFunctionDeclarations.push({
-				arity: paramNames.length,
-				expression: compiledFunctionBody,
-				functionDefinition,
-				localName: declarationLocalName,
-				namespaceURI: declarationNamespaceURI,
-			});
-		}
+		// Only mark the registration as the public API for the module if it's public
```
bwrrp, line 236: This is now also adding private funtions?

src/parsing/processProlog.ts
```diff
@@ -301,12 +315,14 @@ function processFunctionDefinition(
  * @param  prolog            - The AST to process
  * @param  staticContext     - The static context to extend
  * @param  createExpressions - Whether to compile the prolog as well. If we are just parsing there is no need to create expressions already
+ * @param   moduleString      - The string of the module. Used for error reporting if a compilation failed
```
bwrrp, line 318: nit: extra space in front of param name

src/parsing/globalModuleCache.ts
```diff
@@ -1,31 +1,41 @@
...
-) => {
+export const loadModuleFile = (uri: string, moduleContents: ModuleDeclaration) => {
 	let loadedModuleContents = loadedModulesByNamespaceURI[uri];
 	if (!loadedModuleContents) {
 		loadedModuleContents = loadedModulesByNamespaceURI[uri] = {
 			functionDeclarations: [],
 			variableDeclarations: [],
+			performStaticAnalysis: () => {
+				/* By default, do nothing */
+			},
```
bwrrp, line 18: You're using `null` for this elsewhere

src/parsing/processProlog.ts
```diff
@@ -24,6 +29,17 @@ import astHelper, { IAST } from './astHelper';
...
 
+export type ModuleDeclaration = {
+	functionDeclarations: (FunctionDefinition | UpdatingFunctionDefinition)[];
+	source: EvaluableExpression;
+	variableDeclarations: VariableDeclaration[];
+	/**
+	 * Perform static analysis on the module. the whole module is passed, with all additional
+	 * functions that are registered for this namespace
+	 */
+	performStaticAnalysis(moduleContents: ModuleDeclaration): void;
```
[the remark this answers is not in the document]
DrRataplan: dont think so: all calls are non-null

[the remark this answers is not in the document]
bwrrp: the property is set to null [here](https://github.com/FontoXML/fontoxpath/pull/561/files#diff-820513a43476950024f822588211d4f9fc4a64010c96fbf0861e28d2675fe969R100)

src/parsing/processProlog.ts
```diff
@@ -525,7 +542,47 @@ export default function processProlog(
...
+				moduleContents.functionDeclarations.forEach((funDecl) => {
+					if (
+						staticContextLeaf.lookupFunction(
+							funDecl.namespaceURI,
+							funDecl.localName,
+							funDecl.arity,
+							true
+						)
+					) {
+						return;
```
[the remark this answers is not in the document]
DrRataplan: Not really, that's an error somewhere else. It rather prevents duplicate registration of function declared in the local module.

JosVerburg (approved):

src/registerXQueryModule.ts
```diff
@@ -6,13 +6,40 @@ import StaticContext from './expressions/StaticContext';
...
+ * a dependency module are available in a registered module. All public functions and variables
+ * declared in the same namespace as a registered module, but registered in a different
+ * `registerXQueryModule` call are also available.
+ *
+ * @example
+ *
+ * ```
+ *
+ * registerXQueryModule(`module namespace example = "http://www.example.com";
+ * declare %public function example:hello () as xs:string { "hello world" \};
```
bwrrp, line 30: I don't think the `}` needs escaping here?

src/parsing/globalModuleCache.ts
```diff
@@ -57,3 +70,27 @@ export const enhanceStaticContextWithModule = (staticContext: StaticContext, uri
...
+ * @public
+ */
+export const performStaticCompilationOnModules = () => {
+	Object.keys(loadedModulesByNamespaceURI).forEach((namespaceURI) => {
+		const moduleContents = loadedModulesByNamespaceURI[namespaceURI];
+		if (moduleContents.performStaticAnalysis) {
+			try {
+				moduleContents.performStaticAnalysis(moduleContents);
+			} catch (error: unknown) {
+				moduleContents.performStaticAnalysis = null;
```
bwrrp, line 90: What would happen the next time someone tries to call this module? Is it left in a consistent state, or should we put a separate try/catch around each individual registered module's performStaticAnalysis call to make sure valid modules do not skip static analysis if an earlier module in the same namespace fails?

DrRataplan: It throws and then set up for future compilation. A static error may be resolved later

PR #577, opened 2023-02-01
Dramatically improve parse errors

DrRataplan:
* Give a better location by adding some strategic `cut`s in the parser
* Make the error more readable by providing more context


src/parsing/literalParser.ts
```diff
@@ -162,10 +163,12 @@ export const elementContentChar = preceded(
 export const cdataSection: Parser<IAST> = map(
 	delimited(
 		tokens.CDATA_OPEN,
-		star(
-			preceded(
-				peek(not(tokens.CDATA_CLOSE, ['CDataSection content may not contain "]]>"'])),
-				char
+		cut(
```
bwrrp, line 166: `delimited` has a pretty handy fourth argument to automatically insert a `cut` after the opening parser succeeds. Setting that to `true` instead of wrapping the middle parser would also mark failures in the closing delimiter be considered fatal.

src/parsing/literalParser.ts
```diff
@@ -200,23 +203,17 @@ export const dirCommentContents: Parser<string> = map(
...
-				or([tokens.X_UPPER, tokens.X_LOWER]),
-				or([tokens.M_UPPER, tokens.M_LOWER]),
-				or([tokens.L_UPPER, tokens.L_LOWER]),
-				(_a, _b, _c) => []
-			),
-			[]
-		)
-	),
-	name
+// Note: we deviate from the spec here. Processing instruction targets must _always_ be a NCName
```
bwrrp, line 210: It's a bit unfortunate that the grammar only links to the XML spec for this production, but the note in the spec linking to https://www.w3.org/TR/2017/REC-xquery-31-20170321/#parse-note-xml-version seems to suggest that the "Namespaces in XML" spec does apply here. I don't think this is necessarily a deviation from the spec...

test/install-assets.ps1
```diff
@@ -1,9 +1,8 @@
...
-Remove-Item ./test/assets/qt3tests/xqueryx-extracted
-Remove-Item ./test/assets/XQUTS-extracted
+-[Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+-Invoke-WebRequest "https://github.com/LeoWoerteler/XQUTS/archive/master.zip" -Out ./test/assets/XQUTS.zip
+-Expand-Archive ./test/assets/qt3tests/xqueryx.zip -DestinationPath ./test/assets/qt3tests/xqueryx-extracted -Force
+-Move-Item ./test/assets/qt3tests/xqueryx-extracted/xqueryx ./test/assets/qt3tests/xqueryx
+-Expand-Archive ./test/assets/XQUTS.zip -DestinationPath ./test/assets/XQUTS-extracted -Force
+-Move-Item ./test/assets/XQUTS-extracted/XQUTS-master ./test/assets/XQUTS
+-Remove-Item ./test/assets/XQUTS.zip
+-Remove-Item ./test/assets/qt3tests/xqueryx-extracted
```
bwrrp, line 8: Did you mean to insert all of those leading `-`s?

PR #603, opened 2023-08-02
Fix text() not selecting CDATA sections

bwrrp:
While the expression version of the text() node test did check for the CDataSection node type, the codegen-emitted code did not. Additionally, the bucket generated for a CDataSection node was type-4, while the text() selector always uses type-3. This fixes both issues and adds a few tests to confirm that fix.

### DIFF
--- a/src/getBuckets.ts
+++ b/src/getBuckets.ts
@@ -42,7 +42,11 @@ export function getBucketsForPointer(pointer: NodePointer, domFacade: DomFacade)
  * @param node - The node which buckets should be retrieved
  */
 export function getBucketsForNode(node: Node): Bucket[] {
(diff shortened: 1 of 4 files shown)


DrRataplan (approved):

DrRataplan:
Looks great! Thanks! The parser for the QT3 tests just consider CDataSections as textnodes, so that's why this never surfaced. I guess with an upgrade to Slimdom#latest this will kick in tho

bwrrp:
I noticed the XPath playground also seems to convert CData to text, so I could not repro there. If you do want to keep that behavior when switching to the slimdom parser, you can use the `treatCDataAsText` option I added.

MathijsMeijer (approved):

Issue #351, opened 2021-05-21
Crash in math:pow


devatwork:
Found by #349 

Selector: `math:pow(-2.5!0, 2.0e0)`
Language: `XPath3.1`
```
TypeError: Cannot read property 'ta' of null
    at Sg (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:94:62)
    at Object.next (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:93:300)
    at Object.next (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:263:551)
    at Object.next (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:58:985)
    at Xc.E.U (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:59:294)
    at Xc.E.jb (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:58:1179)
    at ok (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:164:419)
    at un.evaluate (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:247:58)
    at Array.<anonymous> (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:177:781)
    at C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:179:76
```

devatwork:
Same error, slightly different stack trace, probably some issue.

!!! Crash found !!!
Selector: `fn:round-half-to-even(3.5678!2<+3, 2)`
Language: `XQueryUpdate3.1`
```
TypeError: Cannot read property 'ta' of null
    at Sg (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:94:62)
    at Object.next (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:93:300)
    at Object.next (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:263:551)
    at Object.next (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:58:985)
    at Xc.E.U (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:59:294)
    at Xc.E.jb (C:\fontoxml\platform\fontoxpath\dist\fontoxpath.esm.js:58:1179)
[...]

PR #606, opened 2023-08-16
Allow map operator to be evaluated statically

DrRataplan:
Without any context. This used to crash with a type error


DrRataplan:
Fixes #351 and maybe more

PR #607, opened 2023-08-16
Format integer RTL

drex04:
Extends the format-integer XPath function to be able to handle numbering styles from Arabic, Hebrew, and Persian.


DrRataplan:
small nitpick, but the spec actually requires one to allow the whole digit family range for that writing system to be allowed:
```
The corresponding output format is a decimal number, using this digit family, with at least as many digits as there are mandatory-digit-signs in the format token. Thus, a format token 1 generates the sequence 0 1 2 ... 10 11 12 ..., and a format token 01 (or equivalently, 00 or 99) generates the sequence 00 01 02 ... 09 10 11 12 ... 99 100 101. A format token of &#x661; (Arabic-Indic digit one) generates the sequence ١ then ٢ then ٣ ...

```

Note that my [favourite reference implementation](https://www.videlibri.de/cgi-bin/xidelcgi) also does not do this correctly, so I'm fine with the current state.

A rebase should fix the bundle size stuff by the way

DrRataplan (changes requested):

PR #608, opened 2023-09-11
Improve stack traces for errors in external functions

DrRataplan:
This should make debugging a bit easier. Just add a stacktrace layer for the function call with the function name


src/expressions/debug/StackTraceGenerator.ts
```diff
@@ -74,7 +101,8 @@ export type Location = {
 	offset: number;
 };
 
-export type SourceRange = {
+export declare type SourceRange = {
```
bwrrp, line 104: Why the declare instead of importing this type where it's used?

DrRataplan: It was because something somewhere was renaming the `comment` field inconsistently. This is now fixed.

bwrrp (approved):

PR #634, opened 2024-04-19
Include Martin in the license for FontoXPath

DrRataplan:
As dicussed with Bert Willems.


devatwork (approved):
It took me a while to get internal approval, but we're good to merge.

PR #639, opened 2024-05-22
Implement the fn:lang function

egh:
- Implemented the [`lang`](https://www.w3.org/TR/xpath-functions-31/#func-lang) function.
- Added some basic tests for this
- Not sure how to properly run the QT3 test suite to test this - I ran the test suite and everything passed, but with pending tests.

### DIFF
--- a/test/assets/runnableTestSets.csv
+++ b/test/assets/runnableTestSets.csv
@@ -111,7 +111,7 @@ fn-insert-before,true
 fn-iri-to-uri,true
 fn-json-doc,false
 fn-json-to-xml,false
-fn-lang,false
+fn-lang,true
 fn-last,true
 fn-load-xquery-module,false
 fn-local-name,false
--- a/test/assets/unrunnableTestCases.csv
+++ b/test/assets/unrunnableTestCases.csv
@@ -607,10 +607,6 @@ fn-function-lookup-259,AssertionError: Expected XPath exists(function-lookup(fn:
 fn-function-lookup-260,Error: XPTY0004: Expected base expression of a function call to evaluate to a sequence of single function item
 fn-function-lookup-261,AssertionError: Expected XPath exists(function-lookup(fn:QName('http://www.w3.org/2005/xpath-functions', 'in-scope-prefixes'), 1)) to resolve to true: expected false to be true
 fn-function-lookup-262,Error: XPTY0004: Expected base expression of a function call to evaluate to a sequence of single function item
-fn-function-lookup-275,AssertionError: Expected XPath exists(function-lookup(fn:QName('http://www.w3.org/2005/xpath-functions', 'lang'), 1)) to resolve to true: expected false to be true
-fn-function-lookup-276,Error: XPTY0004: Expected base expression of a function call to evaluate to a sequence of single function item
-fn-function-lookup-277,AssertionError: Expected XPath exists(function-lookup(fn:QName('http://www.w3.org/2005/xpath-functions', 'lang'), 2)) to resolve to true: expected false to be true
-fn-function-lookup-278,Error: XPTY0004: Expected base expression of a function call to evaluate to a sequence of single function item
 fn-function-lookup-318,Error: FOCH0002: No collations are supported
 fn-function-lookup-320,Error: XPTY0004: eqOp not available for xs:integer and xs:string
 fn-function-lookup-322,Error: FOCH0002: No collations are supported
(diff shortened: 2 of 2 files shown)


DrRataplan:
Hey Erik,

Thanks so much! I'll fix some small things when I get home again, but to answer your questions:

- The QT3 tests can be done using the command `npm run alltests -- --regenerate`. This will empty out the known-failing list and refill it with new errors. It takes some time though, so make sure you can get a coffee while waiting :wink:
- But before that, the fn-lang-related tests should be marked as possibly running. You can do that here: https://github.com/FontoXML/fontoxpath/blob/master/test/assets/runnableTestSets.csv#L114. Just set it to `true` before running the tests.
- I see some small linting / formatting errors. Pesky tabs versus spaces. A `npm run lint-fix` should resolve that.

I'll give it a review when I get home, but looks great! And thanks so much for the tests!

Kindest regards,

Martin

egh:
> Hey Erik,
> 
> Thanks so much! I'll fix some small things when I get home again, but to answer your questions:
> 
>     * The QT3 tests can be done using the command `npm run alltests -- --regenerate`. This will empty out the known-failing list and refill it with new errors. It takes some time though, so make sure you can get a coffee while waiting 😉
> 
>     * But before that, the fn-lang-related tests should be marked as possibly running. You can do that here: https://github.com/FontoXML/fontoxpath/blob/master/test/assets/runnableTestSets.csv#L114. Just set it to `true` before running the tests.
> 
>     * I see some small linting / formatting errors. Pesky tabs versus spaces. A `npm run lint-fix` should resolve that.
> 
> 
> I'll give it a review when I get home, but looks great! And thanks so much for the tests!


I fixed the linting problems, enabled & ran the QT3 tests, and fixed the bugs it revealed. There are 2 outstanding `lang` tests that are failing, but I frankly don't understand the xpath so I'm sure what is going wrong.

DrRataplan:
Hey Erik,

Thanks! The test that is still failing is of the shape `let $l := /langs/para[4]!fn:lang#1 return /langs/para[1]!$l('de')`. I think it's trying to bind the context item of paragraph 4 to the fn:lang function. I'll have to look further into that, but it looks like the function binding is misbehaving here, rather than the lang function.

This does not block merging.

I'll wait until the pipeline is done and I'll merge it!

Thanks again!

Martin

DrRataplan (approved):

src/expressions/functions/builtInFunctions_node.ts
```diff
@@ -165,6 +165,78 @@ function areSameNode(
...
+ * Starting at a given node, test the xml:lang attribute to see if it
+   matches a test. Move up the tree until the test passes or fails.
+ */
+function langMatchNode(node: NodePointer, domFacade: DomFacade, test: string): ISequence {
+	let toTest = node;
+	while (toTest) {
+		if (domFacade.getNodeType(toTest) !== NODE_TYPES.ELEMENT_NODE) {
+			toTest = domFacade.getParentNodePointer(toTest as AttributeNodePointer);
+		} else if (domFacade.getNodeType(toTest) === NODE_TYPES.ELEMENT_NODE) {
+			const attrValue = domFacade.getAttribute(toTest as ElementNodePointer, 'xml:lang');
```
bwrrp, line 194: Should this be a namespace-aware getAttributeNS?

src/expressions/functions/builtInFunctions_node.ts
```diff
@@ -165,6 +165,78 @@ function areSameNode(
...
+/**
+ * Test if a string should match the xml:lang attribute string of a node.
+ */
+function langMatchString(truth: string, test: string): boolean {
+	truth = truth.toLowerCase();
+	test = test.toLowerCase();
+	if (truth === test) {
+		return true;
+	} else {
+		return truth.length < 5 || !truth.startsWith(test)
```
bwrrp, line 177: Are we certain no valid tags are prefixes of each other?
Also, why 5? Maybe checking for the presence of a `-` is a better exit cause for the recursion.

src/expressions/functions/builtInFunctions_node.ts
egh: Weirdly, it's required to use the `xml` prefix, unlike other namespaces. https://www.w3.org/XML/1998/namespace

src/expressions/functions/builtInFunctions_node.ts
egh: The `startswith` is just a heuristic, the real testing is done after splitting off the last part.

`< 5` because it will match `en-us` and won't match `en`. Maybe it should be `truth.contains("-")`

src/expressions/functions/builtInFunctions_node.ts
bwrrp: Ah true, that would make this equivalent 👍

bwrrp (approved):

PR #641, opened 2024-05-31
Implement the fn:generate-id function

egh:
https://www.w3.org/TR/xpath-functions-31/#func-generate-id

Simple implementation, but it seems to work.

### DIFF
--- a/src/expressions/functions/builtInFunctions_node.ts
+++ b/src/expressions/functions/builtInFunctions_node.ts
@@ -30,7 +30,7 @@ import { errXPDY0002 } from '../XPathErrors';
 import { BuiltinDeclarationType } from './builtInFunctions';
 import builtinStringFunctions from './builtInFunctions_string';
 import FunctionDefinitionType from './FunctionDefinitionType';
-
+import generateId from './generateId';
 const fnString = builtinStringFunctions.functions.string;
 
 function contextItemAsFirstArgument(
@@ -121,6 +121,32 @@ const fnData: FunctionDefinitionType = (
 	sequence,
 ) => {
 	return atomize(sequence, executionParameters);
+};
+
+const fnGenerateId: FunctionDefinitionType = (
+	dynamicContext,
+	executionParameters,
+	_staticContext,
+	nodeValue,
+) => {
+	let node: NodePointer;
+	if (!nodeValue) {
+		if (!dynamicContext || !dynamicContext.contextItem) {
+			throw errXPDY0002(
+				`The function generate-id depends on dynamic context if a node is not passed.`,
+			);
+		}
+		if (!isSubtypeOf(dynamicContext.contextItem.type, ValueType.NODE)) {
+			throw new Error('XPTY0004: The context item must be a node.');
+		}
+		node = dynamicContext.contextItem.value;
+	} else {
+		if (nodeValue.isEmpty()) {
+			return sequenceFactory.singleton(createAtomicValue('', ValueType.XSSTRING));
+		}
+		node = nodeValue.first().value;
+	}
+	return sequenceFactory.singleton(createAtomicValue(generateId(node), ValueType.XSSTRING));
 };
 
 const fnHasChildren: FunctionDefinitionType = (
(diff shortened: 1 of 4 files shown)


egh:
@DrRataplan Thank you for looking at this! I wasn't sure if those xpath features were implemented (my xpath knowledge mostly stops at 2.0 :grinning: ). It turns out the counter wasn't properly being passed to some children. I believe I have corrected the issue.

egh:
Looks like there are some type errors I need to fix...

egh:
I just realized that this won't really work for my use case (an xslt processor) unless I allow the caller of an `evaluateXpath` function to pass in a counter. I'll give that a shot...

DrRataplan:
Is that because you need the result to be deterministic over multiple
calls? A weakmap on the module level should fit that, right? Otherwise,
something on DomFacade would not be too dirty...

egh:
> Is that because you need the result to be deterministic over multiple calls? A weakmap on the module level should fit that, right? Otherwise, something on DomFacade would not be too dirty...

Yes, that's right... I think a DomFacade seems like the way to go. Thanks for the tip.

src/expressions/DynamicContext.ts
```diff
@@ -16,6 +16,8 @@ class DynamicContext {
 	public contextItemIndex: number;
 	public contextSequence: ISequence;
 	public variableBindings: { [s: string]: () => ISequence };
+	public generateIdMap: WeakMap<Node, string>;
```
DrRataplan, line 19: If you move this weakmap to the module level, as in, make it its own file like this:

```
const generateIdMap: WeakMap<Node,string> = new WeakMap();
export default generateIdMap;
```

and just directly talk to that in the generate-id function, circumventing the DynamicContext, you will get that stability over multiple calls. Because this is a weakMap I don't expect any memory downsides. I'd like this more than externalizing this state with a DomFacade. That would require other users of FontoXPath to do some upgrading on their side, whcih I'm trying to minimize.

DrRataplan (changes requested):

src/expressions/DynamicContext.ts
egh: Makes sense to me! I'll give this a shot. Thank you

egh: I implemented the function using a module, but the `WeakMap` is not working as I expected with nodes. It doesn't seem to be recognizing previous nodes. I'm trying to track this down, but I would have expected nodes to be equal across subsequent calls. See the failing test here https://github.com/FontoXML/fontoxpath/actions/runs/9475980881/job/26108159273?pr=641#step:6:100

DrRataplan: NodePointers. Forgot about them.

In XQuery (Update Facility) we are sometimes cloning nodes, a lot. When you wrap a large structure (like `let $ele := <ele/> let $wrap := <wrap>{$ele}</wrap> return $wrap`) the `<ele/>` in `$ele` and the one in the `$wrap` are not equal (`$ele is $wrap/ele` is false). [playground](https://xpath.playground.fontoxml.com/?state=H4sIAEQ9aWYAA21Ry2rDMBD8lcUU0kKw6TU4ht5TaEsLKfii2OtYVC%2BkVWoT8u9dyQ05tCDQrDQ7O9KcC217LDaP6%2BIkvBQHhaHYFOdLsS4mrRjWvDWtAahJksJmcoLG0ikxH72Npi8Ha8gyqeysrquFlPkhai383LyPMgAvAQqFN9IcgaxVMFgP%2B%2BfdGvYvLAnC9LB%2Fjejnsq6uvb%2BDXchowSD77Qp7Savm00bohIFUAZ64l8akbw3QiDxvoGTJ%2FemehHb81JuC6DoMAbT1CNdbGLzVbFujiSAXSbIOvDyO%2F%2Bs69Fooab5WzZOxzPdwiESLHVZWSAFmHhlGwSUjD4SBIIZkm318vO1uyhmlp9dVSiFlkj%2BIY2EluEOFsNlCzXvFrHz27YXLhwk058S51FUuWuORojdw35qWcjfHkjsqLlrzwBM6jhMnus4pLj%2BU9WCiIwIAAA%3D%3D)

Figuring out how to circumvent this. The `nodeCompare` function compares node identity, so that's close. We might have to memoize the `createPointer` function in DomFacade.ts here. That is the only place we make NodePointers. This way it can work over XQuery-made nodes as well.

egh: This seems to work for the tests I wrote. There are probably situations where it won't work:

```
import { NodePointer, TinyNode } from '../../domClone/Pointer';
import { ConcreteNode } from 'src/domFacade/ConcreteNode';

const generateIdMap: WeakMap<TinyNode | ConcreteNode, string> = new WeakMap();
let generateIdCounter = 0;

function generateId(ptr: NodePointer): string {
  const node = ptr.node;
	if (!generateIdMap.has(node)) {
		generateIdMap.set(node, `id${++generateIdCounter}`);
	}
	return generateIdMap.get(node);
}

export default generateId;
```

egh:
I'm pretty happy with what I have here - what do you think?

DrRataplan:
I think it looks great! I'll have to take a proper gander next week, if
that is OK, after Midsommar. Thanks again!

egh:
No problem! Thank you. I just wanted to let you know it was ready for your review. Thanks for everything.

src/expressions/functions/builtInFunctions_node.ts
```diff
@@ -123,6 +123,32 @@ const fnData: FunctionDefinitionType = (
...
 
+const fnGenerateId: FunctionDefinitionType = (
+	dynamicContext,
+	executionParameters,
+	_staticContext,
+	nodeValue,
+) => {
+	let node: NodePointer;
+	if (!nodeValue) {
+		if (!dynamicContext || !dynamicContext.contextItem) {
```
DrRataplan, line 134: You could've used the `contextItemAsFirstArgument` helper, that does this internally, but I'm fine either way.

DrRataplan (approved):
LGTM! Let's merge it!

DrRataplan:
Finally got around to those linting errors. Thanks again for the help!

DrRataplan (approved):

DrRataplan:
Since I'm no longer with Fonto I cannot release this. @bwrrp, can you hit the ship button somewhere after your holidays?

egh:
Thank you @DrRataplan ! and best of luck with the new work.

bwrrp:
Sorry for the delay, just released 3.33.0 which includes this

egh:
:pray: Thank you!

PR #659, opened 2024-10-22
Fix demo page

DrRataplan:
Use vite instead of old snowpack. Works a lot easier, just requires some `type` imports in index.ts.
@bwrrp , you're on windows now right? Can you give this a quick run?


bwrrp (approved):
No config needed? Nice!

Works perfectly on WSL, but I had to delete package-lock.json on Windows to work around a bug in npm where it didn't install `@rollup/rollup-win32-x64-msvc` ☹️ 
That was clearly explained in the error though, so approving anyway.

DrRataplan:
Awesome! That sounds scary. But happy to hear it works!

PR #668, opened 2025-03-05
SVG games written using XQuery Update Facility

hfmanson:
SVG games written in XQUF. See description in README


hfmanson:
The XQuery games will move to a separate repository

PR #673, opened 2025-03-19
Fix handling of AST expressions in fontoxpath:evaluate error handling

### LINKED ISSUE #672: Primary error hidden if fontoxpath:evaluate is applied to a precompiled expression that throws
This was a fun one to debug - our built-in `fontoxpath:evaluate` function can be applied to both strings (direct XPath source) or elements (representing XPath ASTs). The latter form is used for Fonto Editor's `xq`. Turns out we have some type confusion around the value here in the error handling logic. This works fine in case the expression is a string, but if it's an element the value is treated as an Element while it is in fact a NodePointer to one. This causes a crash in `printAndRethrowError` as it's trying to iterate over its `.childNodes` while it should have used `.node.childNodes` when trying to find the comment that contains the original expression.

bwrrp:
Fixes #672

### DIFF
--- a/src/expressions/functions/builtInFunctions_fontoxpath.ts
+++ b/src/expressions/functions/builtInFunctions_fontoxpath.ts
@@ -42,7 +42,7 @@ function buildResultIterator(
 	args: ISequence,
 	staticContext: StaticContext,
 	executionParameters: ExecutionParameters,
-): { queryValue: Value; resultIterator: IIterator<Value> } {
+): { evaluableExpression: EvaluableExpression; resultIterator: IIterator<Value> } {
 	const queryValue = query.first();
 	const variables = (args.first() as MapValue).keyValuePairs.reduce((expandedArgs, arg) => {
 		expandedArgs[arg.key.value] = createDoublyIterableSequence(arg.value());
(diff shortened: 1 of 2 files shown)


DrRataplan (approved):

DrRataplan (changes requested):
Kind of missing a unit test here. Can you add one?

DrRataplan (approved):

DrRataplan:
Oof. Hope at least the error is now kind-of readable?

bwrrp:
Yep, full error thrown here is now:

```
1: fontoxpath:evaluate($expr, map{ 'n': 1.2 })
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Inner error:
1: format-integer($n,"A")
   ^^^^^^^^^^^^^^^^^^^^^^
      
Error: XPTY0004 Unable to cast argument of type xs:decimal to type xs:integer? while calling format-integer
  at <functionCallExpr (format-integer)>:1:1 - 1:23
  at <functionCallExpr (fontoxpath:evaluate)>:1:1 - 1:44
```

Without the fix we get:

```
1: fontoxpath:evaluate($expr, map{ 'n': 1.2 })
   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
  at <functionCallExpr (fontoxpath:evaluate)>:1:1 - 1:44
```

mvanbutselaar (approved):

PR #682, opened 2026-01-09
feat: add DocumentFragment support for XPath queries

### LINKED ISSUE #681: Support ShadowRoot (DocumentFragment) as a valid context node in browser environments
Hi,

I am testing fontoxpath in a modern browser environment (Chrome) and I've found that it currently doesn't support ShadowRoot as a contextNode.

In Chrome, a ShadowRoot has a nodeType of 11 (DOCUMENT_FRAGMENT_NODE). Currently, any attempt to evaluate an XPath expression against a ShadowRoot fails, regardless of whether the path is absolute or relative.

For example, the following fails when shadowRoot is passed as the context:

Absolute: //div

Relative: .//div

It appears the engine has strict internal checks (for example, in AbsolutePathExpression.ts and likely within the DomFacade traversal logic) that only recognize DOCUMENT_NODE (9) or ELEMENT_NODE (1) as valid entry points or roots.

Since Shadow DOM is a web standard and ShadowRoot acts as the logical root of a scoped tree, it should be treated as a valid context for XPath evaluation. This would allow fontoxpath to be used effectively with Web Components.

Could you please update the node type validation to include DOCUMENT_FRAGMENT_NODE? This would bring fontoxpath in line with modern browser DOM structures.

Tested on: Chrome (latest), fontoxpath (latest).

xjasonli:
Support DOCUMENT_FRAGMENT_NODE (ShadowRoot) as valid root for XPath expressions. Includes namespace resolution fallback to host element since ShadowRoot.lookupNamespaceURI returns null.

Fix: #681

### DIFF
--- a/src/domFacade/ConcreteNode.ts
+++ b/src/domFacade/ConcreteNode.ts
@@ -34,7 +34,7 @@ export type ConcreteDocumentFragmentNode = DocumentFragment & {
 export type ConcreteProcessingInstructionNode = ProcessingInstruction & {
 	nodeType: NODE_TYPES.PROCESSING_INSTRUCTION_NODE;
 };
-export type ConcreteParentNode = ConcreteElementNode | ConcreteDocumentNode;
+export type ConcreteParentNode = ConcreteElementNode | ConcreteDocumentNode | ConcreteDocumentFragmentNode;
 export type ConcreteChildNode =
 	| ConcreteElementNode
 	| ConcreteTextNode
--- a/src/domFacade/DomFacade.ts
+++ b/src/domFacade/DomFacade.ts
@@ -112,7 +112,7 @@ class DomFacade {
 			? parentNode.childNodes
 			: this._domFacade['getChildNodes'](parentNode, bucket);
 
-		if (parentNode.nodeType === NODE_TYPES.DOCUMENT_NODE) {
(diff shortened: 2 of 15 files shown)


DrRataplan:
Thanks for the PR! Just looking through it. It's a tricky one, since the XQuery DOM has no notion of document fragments.  I do agree fontoxpath can just treat them as document nodes, with multiple children. No harm in that.

A test or two would be really useful here. Slimdom has support for document fragments, which is I think the gist of this PR: supporting document fragments. A test for the sibling axis might be specifically interesting: creating a document fragment with two elements, and addressing the second one from the first with `following-sibling::element()`. Should work.

Can we rename the commits to supporting Document Fragments? I first thought we were implementing going from a shadow host into its shadowroot, exposing the shadowroot as a child of its host. Not sure if we want that.


KR,

Martin

xjasonli:
Hi Martin

Thanks for the suggestions!

I just force-pushed the updates. I added some tests for document fragments, and the sibling axis one is working great. I also renamed the commit to "Support document fragments as root/context node" to keep things clear.

Let me know what you think!

xiaoyu066:
This change is very helpful to me.

DrRataplan (approved):

xjasonli:
Thanks for the review and approval, @DrRataplan ! 

It seems GitHub still requires a review from an account with active write permissions to unblock the merge. Could any maintainer help take a look?

PR #685, opened 2026-05-26
Allow attributes to be passed as arguments to custom functions

egh:
I'm not sure why this error checking was there: from what I can tell it works fine to pass an attribute into a custom function and the code that is called to transform the xpath value into a javascript value seems to handle attributes: https://github.com/egh/fontoxpath/blob/allow-attribute-args-for-functions/src/transformXPathItemToJavascriptObject.ts#L140

### DIFF
--- a/test/specs/parsing/registerCustomXPathFunction.tests.ts
+++ b/test/specs/parsing/registerCustomXPathFunction.tests.ts
@@ -280,13 +280,12 @@ describe('registerCustomXPathFunction', () => {
 		);
 	});
 
-	it('disallows attributes as parameters', () => {
-		chai.assert.throws(
-			() =>
-				evaluateXPathToString('test:custom-function3(//@*)', documentNode, null, null, {
-					namespaceResolver: identityNamespaceResolver,
-				}),
-			'Cannot pass attribute nodes',
+	it('allows attributes as parameters', () => {
+		chai.assert.equal(
+			evaluateXPathToString('test:custom-function3(//@*)', documentNode, null, null, {
+				namespaceResolver: identityNamespaceResolver,
+			}),
+			'someValue',
 		);
 	});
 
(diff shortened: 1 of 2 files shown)


test/specs/parsing/registerCustomXPathFunction.tests.ts
```diff
@@ -280,16 +280,6 @@ describe('registerCustomXPathFunction', () => {
 		);
 	});
 
-	it('disallows attributes as parameters', () => {
-		chai.assert.throws(
```
DrRataplan, line 284: Can we reuse this test as 'does allow attributes as parameters'?

DrRataplan (approved):
Coverage dropped marginally because there are fewer branches. Makes sense.

Let's just reuse the older test as a regression test for the future.

Thanks @egh! Must've been a fun puzzle!

DrRataplan:
I have a vague recollection we had this because something in fontonwas wrong with attributes. Maybe this stems from before we had attribute nodes in blueprints?

test/specs/parsing/registerCustomXPathFunction.tests.ts
egh: Sure, I'll reverse the logic!

egh:
@DrRataplan Thanks for the review. I should have added more context. In xjslt I was trying to process a stylesheet which passed attribute nodes as arguments to the `key` function (https://github.com/ncbi/JATSPreviewStylesheets/blob/master/xslt/main/jats-html.xsl#L3267). This was throwing an error. I worked around it for a while by wrapping `@id` in `string(@id)`. But I decided I should look into the underlying issue. I couldn't figure out a way to work around the issue, so I took a look at the fontoxpath code. I found this check. Once I removed the check, everything worked fine (as far as I can tell). I don't know the fontoxpath code that well, but since the tests seem to pass, I thought it would be fine to submit a PR.

test/specs/parsing/registerCustomXPathFunction.tests.ts
egh: Done and force-pushed changes.

DrRataplan:
@bwrrp do you see a moment to review / merge this PR?

bwrrp (approved):
Apologies for the delay, looks good to me! Merging

egh:
Thank you @bwrrp !

